PR #3, opened 2015-06-08
Adds sbt.io.syntax

eed3si9n:
The idea is to introduce a single import that can be used as a DSL.

* For the modular usage people can initially use `import
sbt.io.syntax._`

```scala
scala> import sbt.io.syntax._
import sbt.io.syntax._

scala> file(".") glob "*.md"
res0: sbt.io.PathFinder = 

   ./Migration.md

scala> file(".") globRecursive "*.scala"
res1: sbt.io.PathFinder = 

   ./io/src/main/scala/sbt/internal/io/Alternative.scala
   ./io/src/main/scala/sbt/internal/io/DeferredWriter.scala
   ...
```

/review @jsuereth, @dwijnand


dwijnand:
LGTM. I like this pattern, and I've been exploring it a lot mysql, but I'm still not sure if there's any fall out from it.

io/src/main/scala/sbt/io/RichURI.scala
```diff
@@ -50,7 +50,9 @@ class RichURI(uri: URI) {
     }
 }
 
-object RichURI {
+object RichURI extends URIExtra
+
+trait URIExtra {
```
jsuereth: seriously?  Why the trait?

io/src/main/scala/sbt/io/IOSyntax.scala
```diff
@@ -0,0 +1,13 @@
...
+  type File = java.io.File
+  type URI = java.net.URI
+  type URL = java.net.URL
+
+  def uri(s: String): URI = new URI(s)
+  def file(s: String): File = new File(s)
+  def url(s: String): URL = new URL(s)
+}
+
+object syntax extends IOSyntax
```
jsuereth: Again, why the trait, and then the IOSyntax?

io/src/main/scala/sbt/io/RichURI.scala
eed3si9n: I'm being consistent with sbt's use of trait in other places like `ProjectExtra` - https://github.com/sbt/sbt/blob/0.13.8/main/src/main/scala/sbt/Project.scala

io/src/main/scala/sbt/io/IOSyntax.scala
eed3si9n: Trait allows us to bake `sbt.io.syntax` cake for immediate usage, while reserving the opportunity to bake `sbt` package object when all modules are ready.

dwijnand: Allows allows other clients of sbt-io to bake these into their own package objects without hassle.

jsuereth: @dwijnand I half understand this, but we're definitely asking for heartache mixing traits + package objects....

io/src/main/scala/sbt/io/RichURI.scala
jsuereth: I'm not sure if that's a *good* style to mimic or not.   I'd hate to expose a public trait without a good reason first.

jsuereth:
To sum up, I like the syntax object/can buy the reasoning.  I don't buy the `URIExtra` logic, nor do I like how it is for `Project` in sbt all that much.

io/src/main/scala/sbt/io/IOSyntax.scala
dwijnand: Indeed it's risky. Much like scalaz providing a scalaz.Scalaz import.

Since this I vaguely remember @retronym saying being too clever in package objects leads to bugs. But that might have been more of an offhand comment in the Scala gitter channel than strong advice to avoid.

It's a tradeoff question. I can be swayed either way.

eed3si9n:
@jsuereth @dwijnand We should come up with a systematic style of how we enrich classes in the coding guideline sbt/website#189.

eed3si9n:
Updated IO's syntax to be compliant to the coding guideline.

io/src/main/scala/sbt/io/syntax.scala
```diff
@@ -0,0 +1,30 @@
...
+  def uri(s: String): URI = new URI(s)
+  def file(s: String): File = new File(s)
+  def url(s: String): URL = new URL(s)
+
+  implicit def fileToRichFile(file: File): RichFile = new RichFile(file)
+  implicit def filesToFinder(cc: Traversable[File]): PathFinder = PathFinder.strict(cc)
+}
+
+import java.io.File
+sealed abstract class IOSyntax0 extends IOSyntax1 {
```
jsuereth: this is a complication I hadn't thought about.  We may need to duplicate these hierarchy levels as well in the sbt package object later....

jsuereth:
This does look a lot better to me.  I would be curious how it affects the sbt main package...

I'm ok merging now.  Little worried about implicit priorities...

PR #7, opened 2015-08-31
org.scala-sbt.io organization

eed3si9n:
/review @jsuereth, @dwijnand

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,7 +14,7 @@ lazy val root = (project in file(".")).
   aggregate(io).
   settings(
     inThisBuild(Seq(
-      organization := "org.scala-sbt",
+      organization := "org.scala-sbt.io",
       version := "0.1.0-SNAPSHOT",
       homepage := Some(url("https://github.com/sbt/io")),
       description := "IO module for sbt",


dwijnand:
I've recently personally switched from giving my projects "com.dwijnand.abc" groupId/organisation, to just "com.dwijnand" given I saw you and Josh and other people do the same.

So I'm surprised here. Why change this?

jsuereth:
+1 on @dwijnand comment...

dwijnand:
Thinking about this more, perhaps the intent here is to distinguish between "io" the module that used to be part of sbt, and versioned with sbt, and "io" the standalone project now.

I think I'm leaning towards preferring to keep the coordinates the same.

eed3si9n:
We put our classes in `sbt.io` package so we can publish `sbt.io.IO` that works along side existing sbt 0.13 code base.

I figured it would be somewhat consistent to make our organization name `org.scala-sbt.io` for `sbt.io` package. This is sort of mirroring the Java tradition of package and organization names in sync. For example sbt-aspectj is under `com.typesafe.sbt` package and `com.typesafe.sbt` organization.

> I've recently personally switched from giving my projects "com.dwijnand.abc" groupId/organisation, to just "com.dwijnand" given I saw you and Josh and other people do the same.

The Scala idiomatic tradition is to completely ignore the organization-package name correspondence and claim top-level name randomly. For my projects:

- `package sbtassembly`, organization: `com.eed3si9n`
- `package treehugger`, organization: `com.eed3si9n`
- `package scalaxb.compiler`, organization: `org.scalaxb`

Seeing this, it might make sense to keep `org.scala-sbt` as is.

Another reference point to consider might be [org.scala-lang.modules](http://search.maven.org/#search%7Cga%7C1%7Corg.scala-lang.modules).

jsuereth:
I'd be ok with `org.scala-sbt.modules` but not `org.scala-sbt.io`.  Call me crazy.

PR #8, opened 2015-09-02
publish test artifacts

eed3si9n:
Publishing test artifacts since other modules need it.

/review @jsuereth, @dwijnand


dwijnand:
LGTM

PR #9, opened 2015-09-04
migrate to scalatest 2.2.4

eed3si9n:
/review @jsuereth, @dwijnand


dwijnand:
LGTM. But, by-the-by, why?

eed3si9n:
Fewer dependencies, which hopefully makes it easier to maintain in the long term.
Feature-wise they seem to be pretty similar.

PR #11, opened 2015-09-18
Enable -Xfatal-warnings, Remove remaining warnings

Duhemm:


dwijnand:
Huh. Those changes don't fail `-Ywarn-value-discard`? I could've sworn they would/do...

Anyways, Travis disagrees with me. :+1:

dwijnand:
I guess I'm getting confused with [NonUnitStatements](https://github.com/puffnfresh/wartremover#nonunitstatements) which I often also have enabled.

PR #12, opened 2015-09-29
Return type annotation + Vector

eed3si9n:
This annotates all public methods' return types.
Uses `Vector` when possible.

/review @Duhemm, @dwijnand, @jsuereth

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Alternative.scala
+++ b/io/src/main/scala/sbt/internal/io/Alternative.scala
@@ -4,10 +4,10 @@ import sbt.io.Alternative
 
 private[sbt] object Alternatives {
   import sbt.io.syntax._
-  final def alternatives[A, B](alts: Seq[A => Option[B]]): A => Option[B] =
+  final def alternatives[A, B](alts: Vector[A => Option[B]]): A => Option[B] =
     alts match {
-      case Seq(f, fs @ _*) => alternative(f) | alternatives(fs)
-      case Seq()           => a => None
+      case Vector(f, fs @ _*) => alternative(f) | alternatives(fs.toVector)
+      case Vector()           => a => None
     }
 
   implicit def alternative[A, B](f: A => Option[B]): Alternative[A, B] =
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -93,23 +93,22 @@ object IO {
     {
       val uri = new URI(uriString)
       assert(uri.getScheme == FileScheme, "Expected protocol to be '" + FileScheme + "' in URI " + uri)
-      if (uri.getAuthority eq null)
-        new File(uri)
+      if (Option(uri.getAuthority).isEmpty) new File(uri)
       else {
         /* https://github.com/sbt/sbt/issues/564
-			* http://blogs.msdn.com/b/ie/archive/2006/12/06/file-uris-in-windows.aspx
-			* http://bugs.sun.com/bugdatabase/view_bug.do?bug_id=5086147
-			* The specific problem here is that `uri` will have a defined authority component for UNC names like //foo/bar/some/path.jar
-			* but the File constructor requires URIs with an undefined authority component.
-			*/
+         * http://blogs.msdn.com/b/ie/archive/2006/12/06/file-uris-in-windows.aspx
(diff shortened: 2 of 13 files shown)


dwijnand:
Some minor comments. LGTM.

io/src/main/scala/sbt/internal/io/Alternative.scala
```diff
@@ -4,10 +4,10 @@ import sbt.io.Alternative
 
 private[sbt] object Alternatives {
   import sbt.io.syntax._
-  final def alternatives[A, B](alts: Seq[A => Option[B]]): A => Option[B] =
+  final def alternatives[A, B](alts: Vector[A => Option[B]]): A => Option[B] =
     alts match {
-      case Seq(f, fs @ _*) => alternative(f) | alternatives(fs)
-      case Seq()           => a => None
+      case Vector(f, fs @ _*) => alternative(f) | alternatives(fs.toVector)
```
jsuereth: Technically, this should be list, since you're decomposing head/tail.

io/src/main/scala/sbt/io/PathMapper.scala
```diff
@@ -71,11 +73,12 @@ abstract class Mapper {
    */
   def resolve(newDirectory: File): FileMap = file => if (file.isAbsolute) None else Some(new File(newDirectory, file.getPath))
 
-  def rebase(oldBases: Iterable[File], newBase: File, zero: FileMap = transparent): FileMap =
+  def rebase(oldBases: Vector[File], newBase: File, zero: FileMap = transparent): FileMap =
```
jsuereth: I still think this one should be Iterable...

io/src/main/scala/sbt/io/PathMapper.scala
```diff
@@ -97,16 +100,16 @@ abstract class Mapper {
    * Selects all descendants of `base` directory and maps them to a path relative to `base`.
    * `base` itself is not included.
    */
-  def allSubpaths(base: File): Traversable[(File, String)] =
+  def allSubpaths(base: File): Vector[(File, String)] =
```
jsuereth: Actually, here we may not want a Vector.  "allSubpaths" may want to be a lazy traversable rather than something which keeps all the file objects in memory.

remember the bugs about "running sbt in `/` cause out of memory exception"?   This is an instance where you don't want vector.

io/src/main/scala/sbt/io/PathMapper.scala
```diff
@@ -97,16 +100,16 @@ abstract class Mapper {
...
     selectSubpaths(base, AllPassFilter)
 
   /**
    * Selects descendants of `base` directory matching `filter` and maps them to a path relative to `base`.
    * `base` itself is not included.
    */
-  def selectSubpaths(base: File, filter: FileFilter): Traversable[(File, String)] =
+  def selectSubpaths(base: File, filter: FileFilter): Vector[(File, String)] =
```
jsuereth: see above comment.

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -373,19 +378,17 @@ object IO {
   }
 
   /** Returns the children of directory `dir` that match `filter` in a non-null array.*/
-  def listFiles(filter: java.io.FileFilter)(dir: File): Array[File] = wrapNull(dir.listFiles(filter))
+  def listFiles(filter: java.io.FileFilter)(dir: File): Vector[File] = wrapNull(dir.listFiles(filter))
```
jsuereth: Array => Vector is definitely not an improvement here.  We actually want to move to something even lazier, if we can.  See comments below.

jsuereth:
So, especially for this API where we have a few bugs that we were unable to fix due to the eager-loading of files, I think moving to Vector is a mistake, and will exacerbate true fixes.   Additionally, we should not use Vector when we're doing head/tail decomposition as that's much faster with List.

eed3si9n:
@jsuereth You're right.
I think expressing lazy loading by picking a subtype of commonly used data structure is too subtle and error-prone. I prefer Cats's approach of being more explicit about laziness.
I think we should copy-pasta Eval.scala sans typeclass instances - https://github.com/non/cats/blob/v0.2.0/core/src/main/scala/cats/Eval.scala

PR #18, opened 2016-05-05
Call getCanonicalPath instead of getAbsolutePath

lloydmeta:
Allows us to relativize when basePath is provided as a relative path.

This is useful when we have to work with "schims" for supporting functionality across different Scala/ScalaJS versions that pull in sources from paths relative to the base directory, like in [sourcecde](https://github.com/lihaoyi/sourcecode/blob/master/build.sbt#L22). Some SBT plugins rely on this function to tell whether or not a file is in a directory or not.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -538,7 +538,7 @@ object IO {
    */
   def relativize(base: File, file: File): Option[String] =
     {
-      val pathString = file.getAbsolutePath
+      val pathString = file.getCanonicalPath
       baseFileString(base) flatMap
         {
           baseString =>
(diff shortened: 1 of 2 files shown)


eed3si9n:
LGTM

gslowikowski:
Hi.
I have a question.

In [scoverage plugin](https://github.com/scoverage/scalac-scoverage-plugin) we have problem with calculating source files paths relative to source root directories they are inside for reporting purposes (see [this](https://github.com/scoverage/scalac-scoverage-plugin/pull/159) and [this](https://github.com/scoverage/scalac-scoverage-plugin/issues/150)).

Our issue is not resolved yet. It's in common code used by SBT, Maven and Gradle plugins, so it's important to have working solution for all contexts. SBT 0.13.x didn't canonize paths and I personally was against canonizing paths in our code because of:
- additional overhead
- showing real paths instead of symlinks

My question is, will SBT 1.0 canonize:
- source root paths,
- source files paths?

I don't know where the changed methods are used. Can I test it somehow?

PR #34, opened 2016-12-30
Make IO.relativize relativize less aggressively

dwijnand:
Specifically, following the scaladoc of relativize, make sure (first) that "base" is not a parent of "file" before trying to relativize.

Prior to this change invoking

```scala
IO.relativize(
  file("/target/scala-2.11/resource_managed/main"),   // base
  file("/src/main/resources/scalac-plugin.xml")       // file
)
```

returned

```scala
Some("../../../../src/main/resources/scalac-plugin.xml")
```

which is incorrect according to the scaladoc of relativize and broke sbt's copyResources implementation.

Created PathMapperSpec to specify and assert the correct behaviour, and as a regression test.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -534,8 +534,10 @@ object IO {
   def relativize(base: File, file: File): Option[String] = {
     val basePath = base.toPath
     val filePath = file.toPath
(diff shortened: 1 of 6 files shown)


dwijnand:
@eed3si9n This should fix `compiler-project/src-dep-plugin` scripted test in my branch.

dwijnand:
Confirmed, with a local build of sbt-io this fixes that test.

eed3si9n (approved):

PR #43, opened 2017-04-09
Add parameters to keep temporary dirs and files

smarter:
This is very useful when trying to debug many things in sbt, like
scripted tests.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -309,14 +309,20 @@ object IO {
 
   /**
    * Creates a temporary directory and provides its location to the given function.  The directory
-   * is deleted after the function returns, unless `keepDirectory` is set to true.
+   * is deleted after the function returns if `keepDirectory` is set to false.
    */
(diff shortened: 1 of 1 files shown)


dwijnand (approved):
Yeah, I've seen code switch to `createTemporaryDirectory` (or mkdirs) for this reason.

LGTM, thanks.

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -309,13 +309,13 @@ object IO {
 
   /**
    * Creates a temporary directory and provides its location to the given function.  The directory
-   * is deleted after the function returns.
+   * is deleted after the function returns, unless `keepDirectory` is set to true.
    */
-  def withTemporaryDirectory[T](action: File => T): T =
+  def withTemporaryDirectory[T](action: File => T, keepDirectory: Boolean = false): T =
```
eed3si9n: Plz add override instead of default param - http://www.scala-sbt.org/0.13/docs/Coding-Guideline.html#Prefer+method+overloading+over+default+parameter+values

eed3si9n (changes requested):

eed3si9n (approved):

PR #50, opened 2017-06-25
Make RichFile a value class

xuwei-k:


dwijnand (approved):

eed3si9n:
Why are we making this a value class?

dwijnand:
It's an enrich-my-type/syntax class, so we can avoid allocation.

eed3si9n:
I get that it tries but I'm not sure if it's worth it for non primitive
types given JITing.
Now you have code gen by the compiler.

dwijnand:
It's a 1 line source code change, that has the effect of transforming

```scala
new RichFile(f).exists
```

into

```scala
RichFile.extension$exists(f)
```

Looks like a good trade-off, to me.

eed3si9n:
I was wondering why sbt/sbt couldn't just bump IO to latest, and realized that this change broke bincompat, so it needs to propagate to all modules first.

dwijnand:
Yes, that is expected.

PR #53, opened 2017-06-28
Add option preserveExecutable to IO.copyFile and IO.copyDirectory

dwijnand:
Fixes sbt/sbt#944

Forward port of
https://github.com/sbt/sbt/commit/1e61b9acf34b7d89a54a5adc021cc37cd3ecd0ef

Part of https://github.com/sbt/sbt/issues/2384

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,6 @@ def baseVersion: String = "1.0.0-M12"
 def commonSettings: Seq[Setting[_]] = Seq(
   scalaVersion := scala212,
   javacOptions in compile ++= Seq("-Xlint", "-Xlint:-serial"),
-  scalacOptions in Compile in compile += "-Xfatal-warnings",
   crossScalaVersions := Seq(scala210, scala211, scala212, scala213),
   // mimaPreviousArtifacts := Set.empty // Set(organization.value %% moduleName.value % "1.0.0")
 )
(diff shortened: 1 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?io=pull/53/head&deploy_to_bintray=false">THIS LINK.</A>

eed3si9n (changes requested):
Default params made it impossible to evolve the old IO. We should learn from that and add overloads here.

dwijnand:
You can evolve by just appending to the end.

These methods have 3 optional booleans, how would you like to overload them?

eed3si9n:
```scala
def copy(sources: Traversable[(File, File)]): Set[File]
def copy(sources: Traversable[(File, File)], options: CopyOption): Set[File]
def copy(sources: Traversable[(File, File)],
      overwrite: Boolean,
      preserveLastModified: Boolean,
      preserveExecutable: Boolean): Set[File] 
```

where CopyOption is defined using Contraband.

typesafe-tools:
To validate this pull request against other sbt modules, visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?io=pull/53/head&deploy_to_bintray=false">THIS LINK.</A>

dwijnand:
@eed3si9n Done, but I'm concerned about the impact it has on users migrating from sbt 0.13 to sbt 1... Lemme know.

eed3si9n (approved):
Perfect

eed3si9n:
There might be some source breakages if ppl relied on the default params, but I think we did the best we can to make this evolvable, so I think it's an okay compromise.

PR #56, opened 2017-07-10
Add a check for pre-normalize'd paths, to relativize

dwijnand:
Fixes sbt/sbt#3286
Fixes sbt/sbt#3276

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -564,7 +564,7 @@ object IO {
   def relativize(base: File, file: File): Option[String] = {
     val basePath = base.toPath
     val filePath = file.toPath
-    if (filePath.normalize() startsWith basePath.normalize()) {
+    if ((filePath startsWith basePath) || (filePath.normalize() startsWith basePath.normalize())) {
       val relativePath = catching(classOf[IllegalArgumentException]) opt (basePath relativize filePath)
       relativePath map (_.toString)
     } else None
--- a/io/src/test/scala/sbt/io/IOSpec.scala
+++ b/io/src/test/scala/sbt/io/IOSpec.scala
@@ -23,4 +23,15 @@ class IOSpec extends FlatSpec with Matchers {
     IO.relativize(rootDir.toFile, nestedFile) shouldBe Some("meh.file")
     IO.relativize(relativeRootDir, nestedFile) shouldBe Some("../../meh.file")
   }
+
+  it should "relativize . dirs" in {
+    val base = new File(".")
+    val file1 = new File("./.git")
+    val file2 = new File(".", ".git")
+    val file3 = new File(base, ".git")
+
+    IO.relativize(base, file1) shouldBe Some(".git")
+    IO.relativize(base, file2) shouldBe Some(".git")
+    IO.relativize(base, file3) shouldBe Some(".git")
+  }
 }


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?io=pull/56/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
Could you explain how this fixes https://github.com/sbt/sbt/issues/3286?

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -564,7 +564,7 @@ object IO {
   def relativize(base: File, file: File): Option[String] = {
     val basePath = base.toPath
     val filePath = file.toPath
-    if (filePath.normalize() startsWith basePath.normalize()) {
+    if ((filePath startsWith basePath) || (filePath.normalize() startsWith basePath.normalize())) {
```
eed3si9n, line 567: Is this happening because JIO Path is unable to handle `.` and `.git` correctly?
What if you have `../foo` and `.git`? Don't they start the same?

1. It might make it easier for me to understand if the toPath and if expression part is factored out to an inner function named `sharesPath`.
2. After resolving the symlinks, maybe you should convert it to an absolute URI instead?

eed3si9n (changes requested):

dwijnand:
Of course.

`FileExamples.files` ([link](https://github.com/sbt/util/blob/v1.0.0-M24/internal/util-complete/src/main/scala/sbt/internal/util/complete/ExampleSource.scala#L52-L53)) lists the child paths of a given base directory and then relativizes in terms of that base directory:

```scala
val directory: File = ...
val childPaths = IO.listFiles(directory).toStream
val prefixedDirectChildPaths = childPaths map { IO.relativize(base, _).get } filter { _ startsWith prefix }
```

Because of the fundamental change to use getCanonicalPath instead of getAbsolutePath in sbt 1's IO in https://github.com/sbt/io/pull/18 (with subsequent fixes https://github.com/sbt/io/pull/33 and https://github.com/sbt/io/pull/34) using a base directory of `"."` doesn't relative its children:

```scala
Welcome to Scala 2.12.2 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).

scala> val base = new java.io.File(".")
base: java.io.File = .

scala> val file = new java.io.File("./.git")
file: java.io.File = ./.git

scala> val basePath = base.toPath
basePath: java.nio.file.Path = .

scala> val filePath = file.toPath
filePath: java.nio.file.Path = ./.git

scala> filePath.normalize() startsWith basePath.normalize()
res0: Boolean = false

scala> basePath.normalize()
res1: java.nio.file.Path =

scala> filePath.normalize()
res2: java.nio.file.Path = .git

scala> basePath relativize filePath
res3: java.nio.file.Path = .git
```

So this change fixes the downstream issue, without breaking any of the existing regression tests for #18/#33/#34.

dwijnand:
@eed3si9n I'm happy to iterate on a better fix for this, but I'm keen for this less-than-ideal fix to ship in RC1, please.

eed3si9n (approved):

eed3si9n:
I was concerned about correctness issue, but I can't think of counter example `foo` and `foobar`, so I guess we can merge.

dwijnand:
I share your concern. This fix isn't satisfying in the least. But using it downstreams seems to solve the reported reproductions. Thanks for the merge.

PR #58, opened 2017-07-15
allow passing in a jar timestamp, sort entries by name

raboof:
This is useful for achieving reproducible builds, where building the same sources with the same settings should result in the same artifact.

Related to https://github.com/sbt/zinc/issues/333

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -9,6 +9,7 @@ import sbt.internal.io.ErrorHandling.translate
 import java.io.{
   BufferedReader,
   ByteArrayOutputStream,
+  BufferedOutputStream,
   BufferedWriter,
   File,
   InputStream,
@@ -21,7 +22,7 @@ import java.nio.charset.Charset
 import java.nio.file.FileSystems
 import java.nio.file.attribute.PosixFilePermissions
 import java.util.Properties
-import java.util.jar.{ Attributes, JarEntry, JarOutputStream, Manifest }
+import java.util.jar.{ Attributes, JarEntry, JarFile, JarOutputStream, Manifest }
 import java.util.zip.{ CRC32, ZipEntry, ZipInputStream, ZipOutputStream }
 import scala.annotation.tailrec
 import scala.collection.immutable.TreeSet
@@ -517,11 +518,14 @@ object IO {
   private def writeZip(sources: Seq[(File, String)], output: ZipOutputStream)(
       createEntry: String => ZipEntry
   ) = {
-    val files = sources.flatMap {
-      case (file, name) => if (file.isFile) (file, normalizeName(name)) :: Nil else Nil
-    }
+    val files = sources
+      .flatMap {
+        case (file, name) => if (file.isFile) (file, normalizeName(name)) :: Nil else Nil
+      }
+      .sortBy {
+        case (_, name) => name
+      }
 
-    val now = System.currentTimeMillis
     // The CRC32 for an empty value, needed to store directories in zip files
     val emptyCRC = new CRC32().getValue()
 
@@ -533,21 +537,23 @@ object IO {
     def makeDirectoryEntry(name: String) = {
       //			log.debug("\tAdding directory " + relativePath + " ...")
       val e = createEntry(name)
-      e setTime now
+      // Leave directory timestamps empty to make the build independent of packaging time
+      e setTime 0
       e setSize 0
       e setMethod ZipEntry.STORED
       e setCrc emptyCRC
       e
     }
 
-    def makeFileEntry(file: File, name: String) = {
+    def makeFileEntry(name: String) = {
       //			log.debug("\tAdding " + file + " as " + name + " ...")
       val e = createEntry(name)
-      e setTime getModifiedTimeOrZero(file)
+      // Leave directory timestamps empty to make the build independent of packaging time
+      e setTime 0
       e
     }
     def addFileEntry(file: File, name: String) = {
-      output putNextEntry makeFileEntry(file, name)
+      output putNextEntry makeFileEntry(name)
       transfer(file, output)
       output.closeEntry()
     }
@@ -589,7 +595,16 @@ object IO {
             val main = mf.getMainAttributes
             if (!main.containsKey(MANIFEST_VERSION))
               main.put(MANIFEST_VERSION, "1.0")
-            (new JarOutputStream(fileOut, mf), "jar")
+
+            val os = new JarOutputStream(fileOut)
+            val e = new ZipEntry(JarFile.MANIFEST_NAME)
+            // Leave manifest timestamp empty to make build repeatable
+            e setTime 0
+            os.putNextEntry(e)
+            mf.write(new BufferedOutputStream(os))
+            os.closeEntry()
+
+            (new JarOutputStream(fileOut), "jar")
           case None => (new ZipOutputStream(fileOut), "zip")
         }
       try { f(zipOut) } finally { zipOut.close }


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?io=pull/58/head&deploy_to_bintray=false">this link.</A>

jvican:
Hey @raboof, thanks for doing this! Let me get back to it in a few days, right now we're preparing RC.

dwijnand:
Hey @raboof, do you have any idea what the possible implications of dropping the timestamps are?

Perhaps as alternative to zeroing them out we should introduce one or more parameters so that it becomes a client's decision whether to use currentTimeMillis/File.lastModified or not. WDYT?

raboof:
@dwijnand I'm not aware of possible implications, and I didn't see any warnings on the [reproducible builds maven plugin](https://zlika.github.io/reproducible-build-maven-plugin/index.html) which also clears the timestamps.

In most cases these timestamps represent the 'time of last compilation' or 'time of packaging', I'm not sure when that'd be interesting information. Also I'm not sure this information is even exposed via the classpath API.

I'd be OK with making this behavior configurable (though of course I'd like the default to be 'cleared' ;) ). The question is how/where to configure this and how to pass that information around. I can give it a shot.

jvican:
I still don't have time to look into this, but my gut feeling is that this could have an implication in Zinc to detect changes in jars as well as scalac (to avoid caching the classpath entries). Need to look deeper, but just throwing some observations out there...

typesafe-tools:
A dbuild validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/102/console">is in progress...</a>

dwijnand:
Yeah, that's my concern too.

typesafe-tools:
[//]: # (Refs: 1.0.x-1.0-pull/58/head-1.0-1.0-1.0.x)
dbuild <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/102/">has checked</A> the following projects against Scala 2.12:

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/1.0.x">1.0.x               </A> | sbt/sbt@2d7ec47b13e02526174f897cca0aef585bd7b128
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/1.0">1.0                 </A> | sbt/zinc@d4d29e8ffeecfa7c6a8e834b4ee54c4bfda9af63
  io                | <A HREF="https://github.com/sbt/io/pull/58">pull/58/head        </A> | sbt/io@b75de5a7058f5fbc1181c32461dd7fc7b346740c
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/1.0">1.0                 </A> | sbt/librarymanagement@0147e0c77578ca4852137a6113bcac4c7adeeb38
  util              | <A HREF="https://github.com/sbt/util/tree/1.0">1.0                 </A> | sbt/util@81277cbf62187eabdc615b7c6d3325e9f664a6e1
  website           | <A HREF="https://github.com/sbt/website/tree/1.0.x">1.0.x               </A> | sbt/website@c5ef5f0dc6cd84c6c6977c09484f56113eaad0d7

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild?sbt=1.0.x&zinc=1.0&io=pull/58/head&librarymanagement=1.0&util=1.0&website=1.0.x&deploy_to_bintray=false">(restart)</A>

raboof:
Aah, those kind of tools could be impacted, yes... that might be hard to deal with. Let's see if we can find an example.

jvican:
The thing I'm most worried about is that today I learned that the JVM caches jars based on the timestamp (https://github.com/scala/bug/issues/10295); which conflicts with this change... if a jar changes completely, and the JVM is caching the jars for the same compiler run, a reference to the previous jar could be maintained and that would cause compilation errors.
  
However, I'm not sure how all these pieces interact with each other. I need some kind of first hand experience because I can certainly ensure that will be the result.
  
@raboof I really appreciate you took this thing up. Would you mind creating a sbt plugin that overrides the sbt task creating a jar and nulls out the last modified time and experiment with it? I think that could give us great insights.

raboof:
Interesting find! https://github.com/scala/bug/issues/10295#issuecomment-307696602 is talking about `java.util.File` though - might that refer to the filesystem timestamp rather than the timestamp inside of the Jar (which this PR is about)?

raboof:
@jvican It looks like `IO.jar()` is only called from the `Package` task, but I'm not sure it'd be possible/easy to override that task... could you point me in the right direction?

jvican:
Right, so this only affects the files inside the zip files. Then I'm less scared. But I'm still positive that testing it should give us more confidence :smile: @raboof.
  
To override this in sbt, have a look at the implementation of `packageBin`, which is the responsible for packaging jars IIRC.
  
Let's check that we have project `A`, and project `B`, which depends on `A`. To check that this patch doesn't do anything dangerous (like affecting incremental compilation), I would override `packageBin` with the sbt syntax and reuse the return type, which is the jar (zip) file. It would be something similar to:
  
```scala
// remember to scope it in A
packageBin := {
  val packaged = packageBin.value
  // Write any timestamps inside the `packaged` file to zero
  // Do more stuff to make the jar reproducible
  ???
}
```
  
Following a similar methodology, you can instrument any task of sbt (which is one of the best features of the tool). If it's not by reusing the return type, it's by copy-pasting the original implementation defined in `Defaults.scala` and then modifying the bits you're interested in.
  
Let me know how it goes.

raboof:
Hmm https://github.com/raboof/scala-reproducible-build-sbt-library-dependency doesn't appear to work even with the timestamps intact :thinking:

jvican:
What do you mean by "doesn't appear to work" @raboof?

raboof:
Sorry that was not too helpful :).

When I `sbt publishLocal` `test-lib` and then start an interactive sbt session in `project-using-test-lib` and `run`, that works. When I then change the code in `test-lib` and `sbt publishLocal` it, `run`ing the `project-using-test-lib` again from the (still running) interactive session I get `[error] error while loading Constants, class file '/home/aengelen/.ivy2/local/net.bzzt/test-lib_2.12/0.1-SNAPSHOT/jars/test-lib_2.12.jar(net/bzzt/Constants.class)' is broken`.

Did we expect this to work? If we didn't expect that to work, can you help come up with examples of things we do expect to work (and want to test it keeps working when we strip timestamps)?

jvican:
But you're using `exportJars` right? I was not expecting that to work. What about generating the jars with `packageBin` and then having a dependency on that? What you want to check is that a change produces a different jar (with the same filestamp) but this jar is detected by Zinc and recompilation is triggered.
  
Note: The error you're seeing is because Scalac caches jars when run from sbt or in resident mode...

retronym:
Here's a related [Maven plugin](https://zlika.github.io/reproducible-build-maven-plugin/) that might serve as inspiration.

raboof:
Right: there are basically 2 ways to make sure there are no in-jar timestamps in results: either not generating the timestamps in the first place, or post-processing the jars to remove the timestamps.

This PR works towards the former, while indeed the [maven plugin](https://zlika.github.io/reproducible-build-maven-plugin/) and the [sbt plugin I created on top of that](https://github.com/raboof/sbt-reproducible-builds) post-process the jars to achieve the same.

For now the post-processing 'scratches my itch', but it would be nicer to solve this 'closer to the source'.

raboof:
I'm afraid I'd need more hand-holding to get this over the finish line, so closing this for now. If anyone wants to reopen/continue, feel free (and I'd be happy to help), but I don't feel qualified to push it forward on my own right now :).

raboof:
(continued in https://github.com/sbt/sbt/pull/5233 and https://github.com/sbt/io/pull/279)

PR #65, opened 2017-07-28
Upgrade to Scala 2.13.0-M2

dwijnand:
Builds on #63


eed3si9n (dismissed):

dwijnand:
This is blocked on having a ScalaTest and ScalaCheck for Scala 2.13.0-M2.

PR #79, opened 2017-10-13
patch a perf bottleneck

fommil:
I did some analysis that looked the same as https://github.com/sbt/zinc/pull/371 then I came up with this patch.

I've spent the last few hours trying to apply this as a monkey patch, but I can't get it to work, so I have absolutely no idea if it improves anything or not.

*UPDATE* this needs to actually calculate the hash. My point is that using NIO should be a lot faster than `FileFilterStream`, which is using 99% of my CPU in the profile (not the calculation of the hash).

### DIFF
--- a/io/src/main/scala/sbt/io/Hash.scala
+++ b/io/src/main/scala/sbt/io/Hash.scala
@@ -5,6 +5,7 @@ package sbt.io
 
 import java.io.{ ByteArrayInputStream, File, InputStream }
 import java.net.{ URI, URL }
+import java.nio.file.Files
 
 object Hash {
   private val BufferSize = 8192
@@ -52,7 +53,7 @@ object Hash {
   def apply(as: Array[Byte]): Array[Byte] = apply(new ByteArrayInputStream(as))
 
   /** Calculates the SHA-1 hash of the given file.*/
-  def apply(file: File): Array[Byte] = Using.fileInputStream(file)(apply)
+  def apply(file: File): Array[Byte] = Files.readAllBytes(file.toPath)
 
   /** Calculates the SHA-1 hash of the given resource.*/
   def apply(url: URL): Array[Byte] = Using.urlInputStream(url)(apply)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/79/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

io/src/main/scala/sbt/io/Hash.scala
```diff
@@ -52,7 +53,7 @@ object Hash {
   def apply(as: Array[Byte]): Array[Byte] = apply(new ByteArrayInputStream(as))
 
   /** Calculates the SHA-1 hash of the given file.*/
-  def apply(file: File): Array[Byte] = Using.fileInputStream(file)(apply)
+  def apply(file: File): Array[Byte] = Files.readAllBytes(file.toPath)
```
dwijnand, line 56: Wat. This reads all the bytes in the file. It doesn't calculate the SHA-1 hash for it..

fommil: well then apply the hash to it afterwards... the slow bit is the `FileFilterStream`

like I said, I have absolutely no way of validating anything about this PR.

dwijnand: Ok.

fommil:
```scala
  def apply(file: File): Array[Byte] = apply(Files.readAllBytes(file.toPath))
```

fommil:
using https://github.com/sbt/launcher/pull/43

(and including the `sbt.boot.properties` from my system's `sbt-launch.jar`) I am able to start up sbt with a command like

```
/usr/lib/jvm/java-8-openjdk/bin/java -Dsbt.launcher.monkey=/home/fommil/Projects/io/io/target/scala-2.12/io_2.12-1.1.0-4cdfe23e77f75495dc363fb99748474bedca8edb.jar -XX:MaxMetaspaceSize=1g -Xss2m -Xms1g -Xmx2g -XX:ReservedCodeCacheSize=128m -XX:+CMSClassUnloadingEnabled -jar /home/fommil/Projects/launcher/target/sbt-launch-1.0.2-SNAPSHOT.jar
```

and get a CPU profile in yourkit.

Unfortunately, it looks like the bottleneck is still there in `FilterInputStream` coming when the `Array[Byte]` is wrapped up. In fact, the performance is worse with this patch (and I'm not sure why that should be the case)

So the actual bottleneck is the digest calculation.

I don't understand why we calculate this again for every file, even if the size and timestamp has not changed.

fommil:
relevant https://stackoverflow.com/questions/16321299/java-nio-vs-non-nio-performance

PR #80, opened 2017-10-14
Hash shortcut proof of concept [DO NOT MERGE]

fommil:
Proof of concept of https://github.com/sbt/zinc/issues/433

This reduces the no-op compile of my project (which is very modest in size) from close to 1 minute to the sbt 0.13 time of 20 seconds.

After this, the bottleneck is almost all in scalafmt https://github.com/lucidsoftware/neo-sbt-scalafmt/issues/54

To try this out, build your own sbt launcher https://github.com/sbt/launcher/pull/43 and start it pointing at the jar built by *this* PR. And vote for https://github.com/sbt/sbt/issues/3638 to make this a lot easier.

### DIFF
--- a/io/src/main/scala/sbt/io/Hash.scala
+++ b/io/src/main/scala/sbt/io/Hash.scala
@@ -52,7 +52,18 @@ object Hash {
   def apply(as: Array[Byte]): Array[Byte] = apply(new ByteArrayInputStream(as))
 
   /** Calculates the SHA-1 hash of the given file.*/
-  def apply(file: File): Array[Byte] = Using.fileInputStream(file)(apply)
+  def apply(file: File): Array[Byte] = {
+    val key = (file, file.length, file.lastModified)
+    cache.get(key) match {
+      case Some(hit) =>
+        hit.clone
+      case None =>
+        val result = Using.fileInputStream(file)(apply)
+        cache += (key -> result.clone)
+        result
+    }
+  }
+  var cache: Map[(File, Long, Long), Array[Byte]] = Map.empty
 
   /** Calculates the SHA-1 hash of the given resource.*/
   def apply(url: URL): Array[Byte] = Using.urlInputStream(url)(apply)
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=1.0.0
+sbt.version=1.0.2


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/80/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

io/src/main/scala/sbt/io/Hash.scala
```diff
@@ -52,7 +52,18 @@ object Hash {
...
+    cache.get(key) match {
+      case Some(hit) =>
+        hit.clone
+      case None =>
+        val result = Using.fileInputStream(file)(apply)
+        cache += (key -> result.clone)
+        result
+    }
+  }
+  var cache: Map[(File, Long, Long), Array[Byte]] = Map.empty
```
yawaramin, line 66: This should be `private`, right? Also could you document the intended properties of the cache key, for posterity.

io/src/main/scala/sbt/io/Hash.scala
```diff
@@ -52,7 +52,18 @@ object Hash {
   def apply(as: Array[Byte]): Array[Byte] = apply(new ByteArrayInputStream(as))
 
   /** Calculates the SHA-1 hash of the given file.*/
-  def apply(file: File): Array[Byte] = Using.fileInputStream(file)(apply)
+  def apply(file: File): Array[Byte] = {
+    val key = (file, file.length, file.lastModified)
```
yawaramin, line 56: I think `file.length` is redundant as a key component. If it changes, `file.lastModified` should logically also change, always. Hence we can use `file -> file.lastModified` as the lookup key with the same properties.

yawaramin (commented):
I'm guessing this is a proof-of-concept because you'd really want to do an LRU-style cache here so it doesn't grow unbounded, right? Anyway I like the idea. Also quite surprised that sbt is actually calculating SHAs all the time.

io/src/main/scala/sbt/io/Hash.scala
fommil: this is just to prove that it's the problem... the real fix should be in zinc (and also do it in parallel)

io/src/main/scala/sbt/io/Hash.scala
fommil: if we use the NIO API you basically get both for free... so why not.

io/src/main/scala/sbt/io/Hash.scala
yawaramin: Ah, I see.

fommil:
@yawaramin it's not just recalculating them once per compile, it recalculates them once per project and per config... which means if you have

- utils
- core -> utils
- core-tests -> core
- core-it -> core-tests

we check every jar that utils depends on FOUR times.

fommil:
looks like we're hitting the zeitgeist ... https://github.com/sbt/zinc/pull/427

Issue #82, opened 2017-11-17
PollingWatchService fails to detect file changes...


dwijnand:
While investigating https://github.com/sbt/sbt/issues/3687 I found that `PollingWatchService` is really broken: it doesn't detect modified files.. o_O

I'll submit a PR that adds a pending test to the test suite.

PR #84, opened 2017-11-21
Wait 1s before testing for file contents change

dwijnand:
Looks like on HFS+ time stamp granularity is only 1 second
https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/APFS_Guide/VolumeFormatComparison/VolumeFormatComparison.html
therefore in order to notice a last modified time change you need to
have elapsed into the next second. We'll do this by sleeping 1 second.

Ref #82

I need to review the downstream issue in sbt to see if #82 should be
closed or not.

/cc @Duhemm


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/84/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

dwijnand:
Oh that's interesting - it failed Travis CI too...

PR #87, opened 2017-11-28
Tweak polling

dwijnand:
This is a subset of https://github.com/sbt/io/pull/84, which should pass and be able to merge.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
+++ b/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
@@ -15,7 +15,7 @@ import scala.concurrent.duration.FiniteDuration
 private[sbt] object SourceModificationWatch {
 
   /**
-   * Checks for modifications on the file system every `delayMillis` milliseconds,
+   * Checks for modifications on the file system every `delay`,
    * until changes are detected or `terminationCondition` evaluates to `true`.
    */
   @tailrec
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/87/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
```diff
@@ -97,12 +97,12 @@ private[sbt] final class WatchState private (
     val newKeys =
       fs.filter(Files.exists(_)).foldLeft(registered) {
         case (ks, d) if Files.isDirectory(d) =>
-          if (registered.contains(d)) ks
+          if (ks.contains(d)) ks
```
eed3si9n, line 100: Is this considered a bug fix or an improvement?
If so, is there an epic issue that you can link to, etc?

dwijnand: I'm not 100% certain it would fix a specific buggy use case, but it's an improvement on the status quo: which is that sometimes you'll double register the same directory, which probably leads to slower performance.

Just from a code standpoint would you not think it's wrong to be using the initial ("zero") value of the `foldLeft` rather than the accumulator?

If you're concerned about this causing regression it can't because it's only _not_ double registering files, it doesn't cause some files to no longer be registered.

eed3si9n: The change itself LGTM, but I'd like you to change the commit message to reflect the nature of this fix, mostly so I can roll it up into a release note later.

eed3si9n (changes requested):

io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
dwijnand: sure, will do

eed3si9n (approved):

PR #89, opened 2017-11-29
Fix #86: prevent concurrent modification of keysWithEvents map

mechkg:

PR #90, opened 2017-11-29
Prevent concurrent modification of keysWithEvents map

### LINKED ISSUE #86: File modification events are getting lost in PollingWatchService
While investigating https://github.com/sbt/sbt/issues/3775 I think I have narrowed the bug down to the interaction between `PollingWatchService.pollEvents` and `PollingWatchService.PollingThread.addEvent`. 

Events often get lost without being properly consumed (https://github.com/sbt/io/issues/82 is probably related). 

**UPDATE**: the conclusion below is wrong, see comments further down.

As far as I understand the only reason that could happen is if some events get added between these two statements: 

https://github.com/sbt/io/blob/e8c9757799bb0b251e5c5aefd96f841428d0eb9a/io/src/main/scala/sbt/io/PollingWatchService.scala#L61-L64

so `thread.keysWithEvents.clear` deletes events that weren't consumed. This should be prevented by the `synchronized` guard.

It is difficult to set up my own modified sbt build to properly test this but at a glance I find this line suspicious:

https://github.com/sbt/io/blob/e8c9757799bb0b251e5c5aefd96f841428d0eb9a/io/src/main/scala/sbt/io/PollingWatchService.scala#L105

I don't understand the purpose of the `notifyAll` call at this point and I suspect that it could break the intended synchronization behaviour.

dwijnand:
Fixup and rebase of @mechkg's #89.

Fixes #86
which means
Fixes sbt/sbt#3687
Fixes sbt/sbt#3695
Fixes sbt/sbt#3775

### DIFF
--- a/io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
+++ b/io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
@@ -21,9 +21,9 @@ abstract class SourceModificationWatchSpec(
 
     IO.write(file, "foo")
 
-    watchTest(parentDir)(pollDelay, maxWait) {
-      IO.write(file, "bar")
-    }
+    // watchTest(parentDir)(pollDelay, maxWait) {
+    //   IO.write(file, "bar")
+    // }
     pending // until fixed https://github.com/sbt/io/issues/82
   }
 
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,5 @@
 import Dependencies._
-// import com.typesafe.tools.mima.core._, ProblemFilters._
+import com.typesafe.tools.mima.core._, ProblemFilters._
 
 def baseVersion: String = "1.1.0"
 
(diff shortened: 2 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/90/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n:
```
[info] PollingWatchServiceSpec:
[info] - should detect modified files *** FAILED ***
[info]   false was not equal to true (SourceModificationWatchSpec.scala:322)
```

Should we be worried?

eed3si9n (approved):

io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
```diff
@@ -21,9 +21,9 @@ abstract class SourceModificationWatchSpec(
 
     IO.write(file, "foo")
 
-    watchTest(parentDir)(pollDelay, maxWait) {
-      IO.write(file, "bar")
-    }
+    // watchTest(parentDir)(pollDelay, maxWait) {
+    //   IO.write(file, "bar")
+    // }
     pending // until fixed https://github.com/sbt/io/issues/82
```
ches, line 27: You wrap it as a block I think 😊 

```scala
pendingUntilFixed {
  watchTest(parentDir)(pollDelay, maxWait) {
    IO.write(file, "bar")
  }
}

dwijnand: Ah! Cheers.

PR #91, opened 2017-12-01
Fix redundant registration issue

eatkins:
I noticed that some paths were being repeatedly registered with the
watch service and tracked it down to this code. If a file in the fs
list is not a directory, then its parent would always get registered,
even if there were multiple files with the same parent. The fix is to
use the accumulated map in the fold instead of the seed and to add both
the parent and the child to the accumulated map.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
+++ b/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
@@ -97,13 +97,16 @@ private[sbt] final class WatchState private (
     val newKeys =
       fs.filter(Files.exists(_)).foldLeft(registered) {
         case (ks, d) if Files.isDirectory(d) =>
-          if (registered.contains(d)) ks
+          if (ks.contains(d)) ks
           else ks + (d -> service.register(d, WatchState.events: _*))
 
         case (ks, f) =>
           val parent = f.getParent
-          if (registered.contains(parent)) ks + (f -> registered(parent))
-          else ks + (f -> service.register(parent, WatchState.events: _*))
+          if (ks.contains(parent)) ks + (f -> ks(parent))
+          else {
+            val key = service.register(parent, WatchState.events: _*)
+            ks ++ Seq(parent -> key, f -> key)
+          }
       }
     withRegistered(newKeys)
   }


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/91/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

dwijnand:
See also https://github.com/sbt/io/pull/87#discussion_r153588811.

eatkins:
Oh, shoot, I didn't notice your pull request which makes a similar change. I still think that the else case that I added in which we register both the parent and the child is needed to avoid creating multiple keys for the parent.

dwijnand:
In the case where I had duplication it didn't seem to be necessary. Can you see if you can find a case where it is, please?

eatkins:
SBT does not walk the directories before it walks the rest of the files. In particular, on my system, children can appear before the parent, which means that we make a new WatchKey for every the same parent for every child that appears before it. I have added logging in a custom build that verifies this. 

I added the following diff locally:
```
diff --git a/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala b/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
index 32207c7..d2e3833 100644
--- a/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
+++ b/io/src/main/scala/sbt/internal/io/SourceModificationWatch.scala
@@ -97,15 +97,23 @@ private[sbt] final class WatchState private (
     val newKeys =
       fs.filter(Files.exists(_)).foldLeft(registered) {
         case (ks, d) if Files.isDirectory(d) =>
-          if (ks.contains(d)) ks
-          else ks + (d -> service.register(d, WatchState.events: _*))
+          if (ks.contains(d)) {
+            println(s"Not registering previously registered directory: $d")
+            ks
+          } else {
+            println(s"registering $d")
+            ks + (d -> service.register(d, WatchState.events: _*))
+          }

         case (ks, f) =>
           val parent = f.getParent
-          if (ks.contains(parent)) ks + (f -> ks(parent))
-          else {
+          if (ks.contains(parent)) {
+            println(s"parent of $f is already registered")
+            ks + (f -> ks(parent))
+          } else {
+            println(s"registering $parent for $f")
             val key = service.register(parent, WatchState.events: _*)
-            ks ++ Seq(parent -> key, f -> key)
+            ks ++ Seq( /*parent -> key, */ f -> key)
           }
       }
     withRegistered(newKeys)
```
and got the following output
without adding the parent -> key map:
```
registering /Users/ethanatkins/work/test/src/main/scala
parent of /Users/ethanatkins/work/test/src/main/scala/Foo.scala is already registered
parent of /Users/ethanatkins/work/test/src/main/scala/Main.scala is already registered
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/mkfiles
registering /Users/ethanatkins/work/test/target
registering /Users/ethanatkins/work/test/project
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/.gitignore
registering /Users/ethanatkins/work/test/.git
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/build.sbt
registering /Users/ethanatkins/work/test/.idea
registering /Users/ethanatkins/work/test/src
registering /Users/ethanatkins/work/test/src/test/scala
parent of /Users/ethanatkins/work/test/src/test/scala/Test.scala is already registered
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/mkfiles
Not registering previously registered directory: /Users/ethanatkins/work/test/target
Not registering previously registered directory: /Users/ethanatkins/work/test/project
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/.gitignore
Not registering previously registered directory: /Users/ethanatkins/work/test/.git
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/build.sbt
Not registering previously registered directory: /Users/ethanatkins/work/test/.idea
Not registering previously registered directory: /Users/ethanatkins/work/test/src
```

with the parent to key map
```
registering /Users/ethanatkins/work/test/src/main/scala
parent of /Users/ethanatkins/work/test/src/main/scala/Foo.scala is already registered
parent of /Users/ethanatkins/work/test/src/main/scala/Main.scala is already registered
registering /Users/ethanatkins/work/test for /Users/ethanatkins/work/test/mkfiles
registering /Users/ethanatkins/work/test/target
registering /Users/ethanatkins/work/test/project
parent of /Users/ethanatkins/work/test/.gitignore is already registered
registering /Users/ethanatkins/work/test/.git
parent of /Users/ethanatkins/work/test/build.sbt is already registered
registering /Users/ethanatkins/work/test/.idea
registering /Users/ethanatkins/work/test/src
registering /Users/ethanatkins/work/test/src/test/scala
parent of /Users/ethanatkins/work/test/src/test/scala/Test.scala is already registered
parent of /Users/ethanatkins/work/test/mkfiles is already registered
Not registering previously registered directory: /Users/ethanatkins/work/test/target
Not registering previously registered directory: /Users/ethanatkins/work/test/project
parent of /Users/ethanatkins/work/test/.gitignore is already registered
Not registering previously registered directory: /Users/ethanatkins/work/test/.git
parent of /Users/ethanatkins/work/test/build.sbt is already registered
Not registering previously registered directory: /Users/ethanatkins/work/test/.idea
Not registering previously registered directory: /Users/ethanatkins/work/test/src
```

Notice that in the former case (where I commented out `parent -> key,`), that the directory `/Users/ethanatkins/work/test` was registered three times. With the `parent -> key` entry, `/Users/ethanatkins/work/test` was only registered once.

eatkins:
Made obsolete by #142

PR #92, opened 2017-12-01
Millisecond-precision file modification timestamps using native code

cunei:
See [[JDK-8177809] File.lastModified() is losing milliseconds (always ends in 000)](https://bugs.openjdk.java.net/browse/JDK-8177809).

The Java methods `lastModified()` and `getLastModifiedTime()`, on many JDKs and operating systems, return a number rounded to the whole second, losing the millisecond part. Since those timestamps are used by sbt to detect file changes, the result is that rapid file changes may go undetected, leading to mysterious errors.
This pull request uses JNA (a native code interface) to replace `lastModified()` with native calls that return timestamps with full millisecond precision, on Linux 64-bit, Linux 32-bit, Windows, and macOS. The new methods are called `getModifiedTime()` and `setModifiedTime()`. The code will fall back to Java if no native support is available, or if the property "sbt.io.jdktimestamps" is defined and is not `"false"`.
A diagnostic method `sbt.io.Milli.getMilliSupportDiagnostic(projectDir: File): Option[String]` is supplied to check whether the filesystem/JDK/OS in use support sub-second modification timestamps. It is intended to be run once after the project is loaded, with something like:
```
sbt.io.Milli.getMilliSupportDiagnostic(projectDir) map log.debug
```
If the test fails, it returns a diagnostic string that will remain available in the logs.

### DIFF
--- a/io/src/main/scala/sbt/io/Milli.scala
+++ b/io/src/main/scala/sbt/io/Milli.scala
@@ -1,0 +1,364 @@
+package sbt.internal.io
+
+import java.lang.UnsupportedOperationException
+import java.io.IOException
+import java.io.FileNotFoundException
+import java.io.File
+import java.util.Date
+import java.nio.ByteBuffer
+import java.nio.ByteOrder
+import java.nio.file.Files
+import java.nio.file.{ Paths => JPaths }
+import java.nio.file.attribute.FileTime
+
+import scala.reflect.{ ClassTag, classTag }
+import scala.collection.JavaConverters._
+
+import com.sun.jna.NativeMapped
+import com.sun.jna.Library
+import com.sun.jna.FromNativeContext
+import com.sun.jna.{ Native => JNANative }
+import com.sun.jna.Platform
+
+import com.sun.jna.platform.win32.Kernel32
+import com.sun.jna.platform.win32.WinNT.GENERIC_READ
+import com.sun.jna.platform.win32.WinNT.FILE_SHARE_READ
+import com.sun.jna.platform.win32.WinNT.FILE_SHARE_WRITE
+import com.sun.jna.platform.win32.WinNT.FILE_WRITE_ATTRIBUTES
+import com.sun.jna.platform.win32.WinNT.OPEN_EXISTING
+import com.sun.jna.platform.win32.WinNT.HANDLE
+import com.sun.jna.platform.win32.WinBase.INVALID_HANDLE_VALUE
+import com.sun.jna.platform.win32.WinBase.FILETIME
+import com.sun.jna.platform.win32.WinError.ERROR_FILE_NOT_FOUND
+
+private abstract class Stat[Time_T](size: Int) extends NativeMapped {
+  val buffer = ByteBuffer.allocate(size).order(ByteOrder.nativeOrder())
+
+  def fromNative(nativeValue: Object, context: FromNativeContext) =
+    throw new UnsupportedOperationException("Not supported.")
+
+  def toNative(): Object = buffer
+  def nativeType(): Class[ByteBuffer] = classOf[ByteBuffer]
+
+  def getModifiedTimeNative: (Time_T, Time_T) // (sec, nsec)
+}
+
+private abstract class StatLong(size: Int, mtimeOffset: Int, mtimensecOffset: Int)
+    extends Stat[Long](size) {
+  def getModifiedTimeNative = (buffer.getLong(mtimeOffset), buffer.getLong(mtimensecOffset))
+}
+private abstract class StatInt(size: Int, mtimeOffset: Int, mtimensecOffset: Int)
+    extends Stat[Int](size) {
+  def getModifiedTimeNative = (buffer.getInt(mtimeOffset), buffer.getInt(mtimensecOffset))
+}
+
+private abstract class Milli {
+  def getModifiedTime(filePath: String): Long
+  def setModifiedTime(filePath: String, mtime: Long): Unit
+  def copyModifiedTime(fromFilePath: String, toFilePath: String): Unit
+}
+
+private abstract class MilliNative[Native] extends Milli {
+  def getModifiedTime(filePath: String): Long =
+    fromNative(getModifiedTimeNative(filePath))
+  def setModifiedTime(filePath: String, mtime: Long): Unit =
+    setModifiedTimeNative(filePath, toNative(mtime))
+
+  // If we have more refined native time information, we can use that
+  // information when copying the modified time, rather than using milliseconds
+  override def copyModifiedTime(fromFilePath: String, toFilePath: String): Unit =
+    setModifiedTimeNative(toFilePath, getModifiedTimeNative(fromFilePath))
+
+  protected def getModifiedTimeNative(filePath: String): Native
+  protected def setModifiedTimeNative(filePath: String, mtimeNative: Native): Unit
+  protected def fromNative(mtimeNative: Native): Long // get milliseconds from native time
+  protected def toNative(mtime: Long): Native // convert milliseconds to native time
+}
+
+private abstract class PosixMilli[Interface <: Posix[Time_T]: ClassTag, Time_T]
+    extends MilliNative[(Time_T, Time_T)] {
+  private val options = Map[String, Object]().asJava
+  protected val AT_FDCWD: Int
+  protected val UTIME_OMIT: Time_T
+  private final val ENOENT = 2
+  protected val libc: Interface =
+    (JNANative.loadLibrary(Platform.C_LIBRARY_NAME,
+                           classTag[Interface].runtimeClass.asInstanceOf[Class[Interface]],
+                           options)): Interface
+  protected def fromLongLong(sec_nsec: (Long, Long)): Long = {
+    val (sec, nsec) = sec_nsec
+    sec * 1000L + (nsec / 1000000L)
+  }
+  protected def toLongLong(mtime: Long): (Long, Long) = {
+    val sec = mtime / 1000L
+    val nsec = (mtime - sec * 1000L) * 1000000L
+    (sec, nsec)
+  }
+  protected def checkedIO[T](filePath: String)(f: => Int) = {
+    if (f != 0) {
+      val errno = JNANative.getLastError()
+      val errStr = libc.strerror(JNANative.getLastError())
+      if (errno == ENOENT)
+        throw new FileNotFoundException(errStr + ": " + filePath)
+      else
+        throw new IOException(errStr + ", file: " + filePath)
+    }
+  }
+}
+
+private abstract class PosixMilliLong[Interface <: Posix[Long]: ClassTag]
+    extends PosixMilli[Interface, Long] {
+  protected def fromNative(sec_nsec: (Long, Long)) = fromLongLong(sec_nsec)
+  protected def toNative(mtime: Long) = toLongLong(mtime)
+  protected def setModifiedTimeNative(filePath: String, mtimeNative: (Long, Long)): Unit = {
+    val (sec, nsec) = mtimeNative
+    val times = new TimeSpec2Long(sec, nsec, UTIME_OMIT)
+    checkedIO(filePath) { libc.utimensat(AT_FDCWD, filePath, times, 0) }
+  }
+}
+private abstract class PosixMilliInt[Interface <: Posix[Int]: ClassTag]
+    extends PosixMilli[Interface, Int] {
+  protected def fromNative(sec_nsec: (Int, Int)) = {
+    val (sec, nsec) = sec_nsec
+    fromLongLong((sec.toLong, nsec.toLong))
+  }
+  protected def toNative(mtime: Long): (Int, Int) = {
+    val (sec, nsec) = toLongLong(mtime)
+    (sec.toInt, nsec.toInt)
+  }
+  protected def setModifiedTimeNative(filePath: String, mtimeNative: (Int, Int)): Unit = {
+    val (sec, nsec) = mtimeNative
+    val times = new TimeSpec2Int(sec, nsec, UTIME_OMIT)
+    checkedIO(filePath) { libc.utimensat(AT_FDCWD, filePath, times, 0) }
+  }
+}
+
+// Cannot use classtag to initialize buffer: JNA needs
+// a nullary constructor. So, extend manually below.
+//
+// Note: technically, tv_nsec is a long, not a time_t.
+// But on Linux32 a long is also 4 bytes, so same size as time_t
+// in all cases.
+//
+private abstract class TimeSpec2[Time_T] extends NativeMapped {
+  val buffer: Array[Time_T]
+  def fromNative(nativeValue: Object, context: FromNativeContext) =
+    throw new UnsupportedOperationException("Not supported.")
+  def toNative(): Object = buffer
+  def nativeType(): Class[Array[Time_T]] = classOf[Array[Time_T]]
+}
+
+private class TimeSpec2Long extends TimeSpec2[Long] {
+  val buffer = new Array[Long](4)
+  def this(sec: Long, nsec: Long, UTIME_OMIT: Long) = {
+    this()
+    buffer(0) = 0L
+    buffer(1) = UTIME_OMIT
+    buffer(2) = sec
+    buffer(3) = nsec
+  }
+}
+
+private class TimeSpec2Int extends TimeSpec2[Int] {
+  val buffer = new Array[Int](4)
+  def this(sec: Int, nsec: Int, UTIME_OMIT: Int) = {
+    this()
+    buffer(0) = 0
+    buffer(1) = UTIME_OMIT
+    buffer(2) = sec
+    buffer(3) = nsec
+  }
+}
+
+private trait Posix[Time_T] {
+  def utimensat(dirfd: Int, filePath: String, times: TimeSpec2[Time_T], flags: Int): Int
+  def strerror(errnum: Int): String
+}
+
+private class Linux64FileStat extends StatLong(144, 88, 96)
+private trait Linux64 extends Library with Posix[Long] {
+  def __xstat64(version: Int, filePath: String, buf: Linux64FileStat): Int
+}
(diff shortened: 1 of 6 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/92/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

cunei:
Please do not merge yet, I will use the pull requests to perform additional validations

io/src/main/scala/sbt/io/Milli.scala
```diff
@@ -0,0 +1,364 @@
+package sbt.io
```
eed3si9n: I think this file should move to `sbt.internal.io` package, and we should provide something like `IO.modifiedTime(file)`, etc in `IO`.

Also please copy what you wrote in the pull request description into this commit's message, so we can read it later.

eed3si9n (changes requested):

io/src/main/scala/sbt/io/Milli.scala
eed3si9n: Also some unit test might be helpful (even though we'll mostly run only Linux one via Travis).

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/294/console">is in progress...</a>

io/src/main/scala/sbt/io/Milli.scala
cunei: Sure, it's already planned. Unit testing is coming up as soon as the main validation passes.

typesafe-tools:
[//]: # (Refs: pull/3789/head-pull/463/head-pull/92/head-pull/189/head-pull/134/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/294/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3789">pull/3789/head      </A> | sbt/sbt@66763f9a542b6fa28938e38dd480261cd0e1457b
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/463">pull/463/head       </A> | sbt/zinc@e809ae51b11e1af3df66e83a16f5c20012203b70
  io                | <A HREF="https://github.com/sbt/io/pull/92">pull/92/head        </A> | sbt/io@04a902a982fdd702e8104d99c71050cb32e20c8a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/189">pull/189/head       </A> | sbt/librarymanagement@bc255105b127cb8857a21528535557565cd96701
  util              | <A HREF="https://github.com/sbt/util/pull/134">pull/134/head       </A> | sbt/util@3b404aecaa8e515a7c8f102241d4ab159c171e91
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3789/head&zinc=pull/463/head&io=pull/92/head&librarymanagement=pull/189/head&util=pull/134/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

ijuma:
The linked ticket states that the Files method works correctly. Is that not true?

eed3si9n:
I added the link to JDK-8177809.

@cunei Could you check this?

cunei:
If you are referring to `Files.getLastModifiedTime()`, the answer is no. On my 16.04 Ubuntu:
```
println(getModifiedTime(filePath))   // native
1512178359814

println(Files.getLastModifiedTime((new File(filePath)).toPath).toMillis)
1512178359000
```

eed3si9n:
@cunei Yup that's the one. Thanks for the checking.

eed3si9n:
If anyone reading this want to try it at home. From sbt shell, run:

```
> eval java.nio.file.Files.getLastModifiedTime(java.nio.file.Paths.get("build.sbt")).toMillis
[info] ans: Long = 1512061937000
```

ijuma:
I checked on macOS with APFS and Java 10:

```scala
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 10-ea).

scala> import java.io.File
import java.io.File

scala> import java.nio.file.Files
import java.nio.file.Files

scala> val f = new File("Downloads")
f: java.io.File = Downloads

scala> Files.getLastModifiedTime(f.toPath).toMillis
res0: Long = 1512376609000

scala> f.lastModified
res1: Long = 1512376609808
```

So, it looks like `Files.getLastModifiedTime` is still broken.

ijuma:
I asked in the nio-dev mailing list and Alan clarified:

> The behavior on Linux should be correct for both methods on JDK 10, except if you build OpenJDK with a very old gcc. The discussion on core-libs-dev was 8uX where different versions of gcc are used.
> 
> Once this code is updated in JDK 10 then the behavior should be consistent. It will also sort out the issue on OS X (which did not support the higher precision timestamps when originally implemented).

cunei:
Travis fails because of compiler-bridge_2.13.0-M2:1.0.5, but the rest seems otherwise ok.

cunei:
The five commits concerning millisecond timestamps should be good to go; a unit test is still missing, but right now I would proceed with the merge in order to unblock Dale; also, there is no practical way to write a unit test that covers the various platforms: by definition I can only test the native calls on the architecture in use at build time. The startup diagnostic is also intended to warn if native timestamps do not work as intended (see https://github.com/sbt/sbt/pull/3789#issuecomment-349175440)
So, I would merge at this point. The process is a bit tricky, since it involves all five modules and each depends on the other. So, it should work like this:
- merge https://github.com/sbt/io/pull/92 against io/1.0.x
- bump io version number in 1.0.x, release.
- merge https://github.com/sbt/util/pull/134 against util/1.0.x
- bump util version number in 1.0.x, bump io dependency version, release.
- merge https://github.com/sbt/librarymanagement/pull/189 against lm/1.0.x
- bump lm version number in 1.0.x, bump io, util dependency versions, release.
- merge https://github.com/sbt/zinc/pull/463 against zinc/1.0.x
- bump zinc version number in 1.0.x, bump io, util, lm dependency versions, release.
- merge https://github.com/sbt/sbt/pull/3789 against sbt/1.0.x
- bump sbt version number in 1.0.x, bump io, util, lm, zinc dependency versions, release.
* Then, proceed to 1.x.
- merge each 1.0.x into 1.x.
- make sure there are no additional `lastModified`, `getLastModifiedTime`, `setModified`, `setLastModifiedTime`, or `copyLastModified` calls in 1.x with respect to 1.0.x, and in case replace them with `getModifiedTime`/`setModifiedTime`/`copyModifiedTime`. A tricky bit is that in `internal/zinc-core/src/main/scala/sbt/internal/inc/Stamp.scala`, `util-cache/src/main/scala/sbt/util/FileInfo.scala` there are occurrencies of the `lastModified` identifier which are distinct from the `java.io.File` one, and should be kept as they are.
- re-release the whole chain for 1.x, as above.
I have never released sbt's modules, so I would defer the process to @dwijnand or @eed3si9n. Kindly review the five pull requests and merge as appropriate.

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.0.0
+sbt.version=1.0.4
```
eed3si9n: sbt 1.0.4 is unable to build sbt 2.13.0-M1, so could you exclude it from Travis CI yml?

eed3si9n (dismissed):

project/build.properties
cunei: Right, that's the issue. I'll drop the commit that updates sbt.version.

ijuma:
Btw, is this a significant enough issue that it's worth keeping native code around even though it has been fixed in the JDK itself?

cunei:
The fix is scheduled for inclusion only in JDK 10, but we will need to support JDK 8 and JDK 9 for many years to come. Also, this issue caused some of our tests to fail non-deterministically, and caused a bunch of other collateral issues. So, yep, we really ought to fix this.

ijuma:
JDK 9 will be EOL'd by Oracle in March when JDK 10 ships (it's not a LTS release), so I'd be surprised if anyone supports it for a long time. JDK 8 will be around for a while, I agree.

However, people have the option of running SBT with JDK 10+ while supporting older JDK versions via the `--release` flag (there are plans for scalac to support this flag too) so there won't be a strong reason to run SBT with Java 8 as far as I can tell.

Anyway, you are the ones maintaining the code so your call. :)

eed3si9n:
We came across this behavior during integration testing, but we also suspect it's been affecting other areas in sbt such as triggered execution and incremental compilation.

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/299/console">is in progress...</a>

eed3si9n:
@cunei Thanks for all the ripple PRs. I'm going to run Dbuild and see what happens.

typesafe-tools:
[//]: # (Refs: pull/3789/head-pull/463/head-pull/92/head-pull/189/head-pull/134/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/299/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3789">pull/3789/head      </A> | sbt/sbt@32846825920e5a79a8a209d784d10711b0d2b1fc
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/463">pull/463/head       </A> | sbt/zinc@5d07d2994db55efa325e12bd90556cea86a31804
  io                | <A HREF="https://github.com/sbt/io/pull/92">pull/92/head        </A> | sbt/io@e9400ca4cf2ff9f748f8ce60415270302ff88b2e
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/189">pull/189/head       </A> | sbt/librarymanagement@bc255105b127cb8857a21528535557565cd96701
  util              | <A HREF="https://github.com/sbt/util/pull/134">pull/134/head       </A> | sbt/util@35cd61c562915de7422d1510273c7b537e42ef27
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3789/head&zinc=pull/463/head&io=pull/92/head&librarymanagement=pull/189/head&util=pull/134/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/301/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/3789/head-pull/463/head-pull/92/head-pull/189/head-pull/134/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/301/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3789">pull/3789/head      </A> | sbt/sbt@32846825920e5a79a8a209d784d10711b0d2b1fc
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/463">pull/463/head       </A> | sbt/zinc@5d07d2994db55efa325e12bd90556cea86a31804
  io                | <A HREF="https://github.com/sbt/io/pull/92">pull/92/head        </A> | sbt/io@07ef05614c769f9ddea6ad8eb783eb59a416b79a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/189">pull/189/head       </A> | sbt/librarymanagement@74380f09b655c2653fbd7888600c6585d94e8959
  util              | <A HREF="https://github.com/sbt/util/pull/134">pull/134/head       </A> | sbt/util@35cd61c562915de7422d1510273c7b537e42ef27
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3789/head&zinc=pull/463/head&io=pull/92/head&librarymanagement=pull/189/head&util=pull/134/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

eed3si9n:
As per discussion during sbt 1 meeting, I am retargeting this to 1.1.x branch.

eed3si9n:
The dbuild validation ran up until the vscode stuff, which is not yet in 1.0.x so I'd consider it a pass.

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/302/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/3789/head-pull/463/head-pull/92/head-pull/189/head-pull/134/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/302/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3789">pull/3789/head      </A> | sbt/sbt@a0c884deda2d9330fd549d421438210b5a473833
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/463">pull/463/head       </A> | sbt/zinc@422340c8e0abb31a0cc42c9e3ff3c726498c8028
  io                | <A HREF="https://github.com/sbt/io/pull/92">pull/92/head        </A> | sbt/io@e433a5c62aa94dea4a1203c2c29af97e0d396afc
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/189">pull/189/head       </A> | sbt/librarymanagement@aacab10c79abbbe206ab7b6e251a6ab8e6313fdd
  util              | <A HREF="https://github.com/sbt/util/pull/134">pull/134/head       </A> | sbt/util@d03dfb39817be24a84556cabfc4affc4ac257bb2
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

:white_check_mark: The result is: ***SUCCESS***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3789/head&zinc=pull/463/head&io=pull/92/head&librarymanagement=pull/189/head&util=pull/134/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

cunei:
I temporarily used as base `1.x` for util and lm as well, since `1.1.x` had not been created yet.
The validator has compiled all the branches successfully, but using the result binaries to re-compile sbt failed. It seems there are more points in the code where sbt tries to get timestamps of non-existent files. Here is the run: https://jenkins.scala-sbt.org:8592/job/sbt-community/90/console
```
[sbt-zinc] [error] java.io.FileNotFoundException: No such file or directory: /localhome/jenkinssbt/workspace/sbt-community/target-0.9.9/extraction/76f69ff08fa3644f09af512f3ac4b1ba8edd8ec1/projects/698b3541df4104f83b9b3935c97ecdb2d797dbc7/project/project/target/scala-2.12/sbt-1.0/classes
[sbt-zinc] [error] 	at sbt.internal.io.PosixMilli.checkedIO(Milli.scala:102)
[sbt-zinc] [error] 	at sbt.internal.io.Linux64Milli$.getModifiedTimeNative(Milli.scala:187)
[sbt-zinc] [error] 	at sbt.internal.io.Linux64Milli$.getModifiedTimeNative(Milli.scala:182)
[sbt-zinc] [error] 	at sbt.internal.io.MilliNative.getModifiedTime(Milli.scala:63)
```
I can keep bypassing the exception, but inspecting files that do not actually exist is probably not a great idea; I'd rather check why that happens, and fix the client code instead. OK? @eed3si9n

typesafe-tools:
[//]: # (Refs: pull/3789/head-pull/463/head-pull/92/head-pull/189/head-pull/134/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/303/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3789">pull/3789/head      </A> | sbt/sbt@a0c884deda2d9330fd549d421438210b5a473833
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/463">pull/463/head       </A> | sbt/zinc@422340c8e0abb31a0cc42c9e3ff3c726498c8028
  io                | <A HREF="https://github.com/sbt/io/pull/92">pull/92/head        </A> | sbt/io@e433a5c62aa94dea4a1203c2c29af97e0d396afc
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/189">pull/189/head       </A> | sbt/librarymanagement@aacab10c79abbbe206ab7b6e251a6ab8e6313fdd
  util              | <A HREF="https://github.com/sbt/util/pull/134">pull/134/head       </A> | sbt/util@d03dfb39817be24a84556cabfc4affc4ac257bb2
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3789/head&zinc=pull/463/head&io=pull/92/head&librarymanagement=pull/189/head&util=pull/134/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

fommil:
Cool! However I recommend having a pure jvm fallback in case of exception. Your detection logic could fail, or abis could change, then we're basically not on the jvm anymore.

fommil:
Also, are you sure JNA isn't adding a performance bottleneck? These are very frequent calls and even submilli slowdowns will become apparent on large codebase as multisecond slowdowns. We really need JNI bulk operations to avoid problems. But does it also suffer from precision problems? I'd personally prefer performance over precision in this case, note that we cannot say "correctness" because we're still limited by the filesystem precision.

PR #93, opened 2017-12-06
Small change in Windows millisecond code

cunei:
In Windows, the CreateFile call will fail on directories unless
the special flag FILE_FLAG_BACKUP_SEMANTICS is specified,
because, well, Windows. Works ok for getting/setting the modification
time on directories as well, now.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/93/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

PR #102, opened 2017-12-14
Wait 1s before testing for file contents change

### LINKED ISSUE #82: PollingWatchService fails to detect file changes...

dwijnand:
Looks like on some filesystems, like HFS+, time stamp granularity is only 1 second
https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/APFS_Guide/VolumeFormatComparison/VolumeFormatComparison.html
therefore in order to notice a last modified time change you need to
have elapsed into the next second. We'll do this by sleeping 1 second.

Fixes #82

### DIFF
--- a/io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
+++ b/io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
@@ -16,7 +16,7 @@ abstract class SourceModificationWatchSpec(
 ) extends FlatSpec
     with Matchers {
 
-  it should "detect modified files" in IO.withTemporaryDirectory { dir =>
+  it should "detect modified files" in withTestDirectory { dir =>
     val parentDir = dir / "src" / "watchme"
     val file = parentDir / "Foo.scala"
 
@@ -34,7 +34,7 @@ abstract class SourceModificationWatchSpec(
     }
   }
 
-  it should "watch a directory for file creation" in IO.withTemporaryDirectory { dir =>
+  it should "watch a directory for file creation" in withTestDirectory { dir =>
     val parentDir = dir / "src" / "watchme"
     val created = parentDir / "NewSource.scala"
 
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/102/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n:
```
[info] - should detect modified files *** FAILED ***
[info]   false was not equal to true (SourceModificationWatchSpec.scala:320)
```

dwijnand:
Damn... it failed on Travis CI like it had before: https://github.com/sbt/io/pull/84#issuecomment-346317581

I was hoping @cunei's Milli fix would've fix this.. @cunei any ideas?

io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
```diff
@@ -319,4 +326,17 @@ abstract class SourceModificationWatchSpec(
     WatchState.empty(service, sources).withCount(1)
   }
 
+  private def supportsSubSecondModifiedTime(dir: File) = {
+    val file = File.createTempFile("", "", dir)
```
cunei: Travis says "Prefix string too short"

dwijnand: 🙄

io/src/test/scala/sbt/internal/io/PollingWatchServiceSpec.scala
```diff
@@ -1,12 +1,13 @@
...
 import scala.concurrent.duration._
 
 class PollingWatchServiceSpec
     extends SourceModificationWatchSpec(
       new PollingWatchService(500.milliseconds),
       500.milliseconds,
       3.seconds
-    )
+    ) {
+  Milli.getMilliSupportDiagnostic(IO.temporaryDirectory) map println
```
cunei: This should really be the project dir; the temporary directory on *nix is typically on /tmp, which is tmpfs, a virtual ram-based file system. So, it's not really checking the file system at all.

cunei: Same for the `createTempFile()`, below

dwijnand: Good point..

dwijnand:
This is now green, so it's ready for review.

eed3si9n (approved):

PR #104, opened 2017-12-15
Change PathFinder#get to get()

### LINKED ISSUE #36: Change PathFinder's get to get()
[`PathFinder#get`](https://github.com/sbt/io/blob/v1.0.0-M9/io/src/main/scala/sbt/io/Path.scala#L175) is defined as `def get: Seq[File]`, but its behaviour is to look at the filesystem.

Therefore I argue it should be `def get()`.

With regards to source compatibility fortunately going in this direction is safe.

dwijnand:
Given it side-effects by looking at the FS it should have parens.

With regards to source compatibility fortunately going in this direction
is safe.

Fixes #36

### DIFF
--- a/io/src/main/scala/sbt/io/Path.scala
+++ b/io/src/main/scala/sbt/io/Path.scala
@@ -357,7 +357,7 @@ sealed abstract class PathFinder {
    */
   def pair[T](mapper: File => Option[T], errorIfNone: Boolean = true): Seq[(File, T)] = {
     val apply = if (errorIfNone) mapper | fail else mapper
-    for (file <- get; mapped <- apply(file)) yield file -> mapped
+    for (file <- get(); mapped <- apply(file)) yield file -> mapped
   }
 
   /**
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/104/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

io/src/main/scala/sbt/io/Path.scala
```diff
@@ -375,7 +375,7 @@ sealed abstract class PathFinder {
    * The files returned by this method will reflect the underlying filesystem at the time of calling.
    * If the filesystem changes, two calls to this method might be different.
    */
-  final def get: Seq[File] = {
+  final def get(): Seq[File] = {
```
eed3si9n, line 377: Are we adding empty parentheses because it forces evaluation?

eed3si9n: Would this logic apply to `getURLs`?

dwijnand: Yes to both. I'll look at the uses of get and see if there are more to add.

eed3si9n (approved):

PR #106, opened 2017-12-18
Adds `IO.getModifiedTimeOrZero(file)` and `IO.setModifiedTimeOrFalse(file)`

cunei:
Ref https://github.com/sbt/io/pull/92

There are just too many instances in which sbt's code relies on
the `lastModified`/`setLastModified` semantics, so instead of moving
to `get`/`setModifiedTime`, we use new IO calls that offer the new
timestamp precision, but retain the old semantics.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -1139,7 +1139,7 @@ object IO {
    * @see setModifiedTime
    * @see copyModifiedTime
    */
-  @deprecated
+  @deprecated("This method might be removed in the future, also see getModifiedOrZero()", "1.1.3")
   def getModifiedTime(file: File): Long = Milli.getModifiedTime(file)
 
   /**
@@ -1165,7 +1165,8 @@ object IO {
    * @see getModifiedTime
    * @see copyModifiedTime
    */
-  @deprecated
+  @deprecated("This method might be removed in the future, also see setModifiedTimeOrFalse()",
+              "1.1.3")
   def setModifiedTime(file: File, mtime: Long): Unit = Milli.setModifiedTime(file, mtime)
 
   /**
@@ -1190,7 +1191,7 @@ object IO {
    * @see getModifiedTime
    * @see setModifiedTime
    */
-  @deprecated
+  @deprecated("This method might be removed in the future, also see copyLastModified()", "1.1.3")
   def copyModifiedTime(fromFile: File, toFile: File): Unit =
     Milli.copyModifiedTime(fromFile, toFile)
 
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -306,7 +306,7 @@ object IO {
             }
           }
           if (preserveLastModified)
-            setLastModified(target, entry.getTime)
+            setModifiedTimeOrFalse(target, entry.getTime)
         } else {
           //log.debug("Ignoring zip entry '" + name + "'")
         }
@@ -543,7 +543,7 @@ object IO {
     def makeFileEntry(file: File, name: String) = {
       //			log.debug("\tAdding " + file + " as " + name + " ...")
       val e = createEntry(name)
-      e setTime lastModified(file)
+      e setTime getModifiedTimeOrZero(file)
       e
     }
     def addFileEntry(file: File, name: String) = {
@@ -643,7 +643,7 @@ object IO {
       preserveLastModified: Boolean,
       preserveExecutable: Boolean
   )(from: File, to: File): File = {
-    if (overwrite || !to.exists || lastModified(from) > lastModified(to)) {
+    if (overwrite || !to.exists || getModifiedTimeOrZero(from) > getModifiedTimeOrZero(to)) {
       if (from.isDirectory)
         createDirectory(to)
       else {
@@ -1130,9 +1130,16 @@ object IO {
    * and Java's get/setLastModifiedTime() will be used instead. This
    * setting applies to setModifiedTime() and copyModifiedTime() as well.
    *
+   * This method was added in sbt/io v1.1.2, but it may be
+   * replaced in the future by a similar method that returns
+   * a Try, rather than throwing an exception.
+   * It is therefore marked as deprecated, since we cannot
+   * guarantee that it will remain available in future versions.
+   *
    * @see setModifiedTime
    * @see copyModifiedTime
    */
+  @deprecated
   def getModifiedTime(file: File): Long = Milli.getModifiedTime(file)
 
   /**
@@ -1149,9 +1156,16 @@ object IO {
    *
    * This method may not work correctly if mtime is negative.
    *
+   * This method was added in sbt/io v1.1.2, but it may be
+   * replaced in the future by a similar method that returns
+   * a Try, rather than throwing an exception.
+   * It is therefore marked as deprecated, since we cannot
+   * guarantee that it will remain available in future versions.
+   *
    * @see getModifiedTime
    * @see copyModifiedTime
    */
+  @deprecated
   def setModifiedTime(file: File, mtime: Long): Unit = Milli.setModifiedTime(file, mtime)
 
   /**
@@ -1167,24 +1181,32 @@ object IO {
    * If the timestamp cannot be copied, this method will throw
    * a FileNotFoundException or an IOException, as appropriate.
    *
+   * This method was added in sbt/io v1.1.2, but it may be
+   * replaced in the future by a similar method that returns
+   * a Try, rather than throwing an exception.
+   * It is therefore marked as deprecated, since we cannot
+   * guarantee that it will remain available in future versions.
+   *
    * @see getModifiedTime
    * @see setModifiedTime
    */
+  @deprecated
   def copyModifiedTime(fromFile: File, toFile: File): Unit =
     Milli.copyModifiedTime(fromFile, toFile)
 
   /**
    * Return the last modification timestamp of the specified file,
    * in milliseconds since the Unix epoch (January 1, 1970 UTC).
-   * Please see IO.getModifiedTime() for further information.
    *
-   * In contrast to getModifiedTime(), if the specified file does
-   * not exist, this method will return 0L, rather than throwing
-   * a FileNotFoundException.
+   * If the specified file does not exist, this method will return 0L.
+   *
+   * The deprecated method getModifiedTime() has similar semantics,
+   * but will throw an exception if the file does not exist.
+   * Please refer to its documentation for additional details.
    *
    * @see getModifiedTime
    */
-  def lastModified(file: File): Long =
+  def getModifiedTimeOrZero(file: File): Long =
     try {
       getModifiedTime(file)
     } catch {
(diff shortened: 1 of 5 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?io=pull/106/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -1197,4 +1172,67 @@ object IO {
...
+   * in milliseconds since the Unix epoch (January 1, 1970 UTC).
+   * Please see IO.getModifiedTime() for further information.
+   *
+   * In contrast to getModifiedTime(), if the specified file does
+   * not exist, this method will return 0L, rather than throwing
+   * a FileNotFoundException.
+   *
+   * @see getModifiedTime
+   */
+  def lastModified(file: File): Long =
```
eed3si9n: I think it's too subtle/confusing to have different variety of `getModifiedTime` that differs in similar sounding name.

My suggestion here would be to have two versions:
- `getModifiedTimeTry` that returns `scala.util.Try`.
- `getModifiedTime` that returns `Long` and uses `0L` when the file doesn't exist.

eed3si9n (changes requested):

io/src/main/scala/sbt/io/IO.scala
cunei: `IO.lastModified` and `IO.setLastModified` have the exact same semantics as `f.lastModified` and `f.setLastModified`, which is why retaining the same names is important.
The other methods are just ancillary, and after these patches they will be mostly unused within sbt.
Actually, we have spent a great deal of time already on what should have been a minor matter, so I'd avoid further tweaks. At this point I would strongly recommend that we just merge these pull requests, and move forward with other topics.

eed3si9n: Any public method on `IO` becomes semi-permanent API of sbt, so if we can avoid creating confusing duplicates that will save our headaches in the future. Since io 1.1.2 is already out with `IO.getModifiedTime(..)` we are committed with the name. For the same reason we are going through this exercise, anyone who tries to replace the usage of `File#lastModified` should use the return-`0L` version.

Given all other modules are already calling `IO.getModifiedTime(..)`, changing here also saves changing the code downstream.

cunei: Since `IO.getModifiedTime` is already included in `IO`, we have to keep respecting the specification described in the Scaladoc. It is already described as a method that throws exception, so we cannot change it again now.

If you are concerned about too many time-related methods being exposed in `IO`, it is perfectly ok to only expose `xxxModifiedTime`, keeping the `xxxlastModified` calls internal. I can just place them in `internal.io.Milli`. However, we cannot change `IO.getModifiedTime` again, now.

Consequently, the only possible course of action is transitioning the downstream code to (differently named) calls that go back to the `0L` semantics. That is what these pull requests do, upon @dwijnand's suggestion.

It is convenient to name them in the same way as the `java.util.File` ones, for clarity. It is true however that they do not need to be exposed in `IO` at all. (Incidentally, you are the one who asked for `getModifiedTime` to be exposed in `IO`)

So, as a middle ground solution, I propose that I remove `lastModified` and `setLastModified` from `IO` and move them to `internal.io.Milli`, changing downstream accordingly.

eed3si9n: Given that sbt 1.1.0 hasn't reached final, I am personally ok with `IO.getModifiedTime` changing the semantics from the previous description in Scaladoc.

I am fine with either:
- Change to return `IO.getModifiedTime` `0L`, or
- Make `xxxLastModified` internal.

@dwijnand What do you think?

dwijnand: Given this is sbt/io and not sbt I'm of the opinion that we can't change semantics between stable releases (aka from 1.1.2 to 1.1.3).

What's a good alternative name to `getModifiedTime` to put on `IO` that means "last modified time or 0`. `getLastModifiedOrZero`? `getLastModifiedSafe`?

Personally I'm fine with `lastModified`, but I'd be equally happy if a better name is found that's distinct enough from `getModifiedTime` to avoid confusion.

eed3si9n: How about we deprecate `IO.getModifiedTime` and add:

1. `getLastModifiedTry` that uses `Try`
2. `getLastModifiedOrZero` that returns `0L`

?

dwijnand: Fine by me.

cunei: Then you end up with *three* variants. Confusion is increased even more!... :smile:
We need the zero-returning variant, and we need to keep the exception-throwing variant since it's in the API. There is really no need to add the "Try" one as well...

eed3si9n: If it's deprecated then it's effectively out of the API.
`Try` should be the replacement for someone who wants throwy behavior.

dwijnand: Maybe instead of a `getLastModifiedTry` it should be `getLastModifiedUnsafe` which throws exceptions.

cunei:
We had a long discussion concerning these pull requests, and developed a compromise that, although not particularly pretty in the interim, will eventually get us to a single set of calls with the `Try` semantics. In short:
- The zero-returning variants are problematic, and we will move away from them eventually, but not now
- The exception-throwing variants are better, but having the same prototype as the ones above may be confusing
- The `Try` version is preferable, but we cannot add it right away, otherwise we end up with too many calls in `IO` (get/set/copy, in three variants, would mean having nine different calls in `IO`)

So:

- We are going to stick with the zero-returning semantics for now (equivalent to `java.io.File`'s `lastModified`/`setLastModified`)
- In order to make it clear that the methods may deal with zero timestamps, however, in `IO` we will use for them the names `getModifiedTimeOrZero()` and `setModifiedTimeOrFalse()` (the existing `copyLastModified()`, already in the API, will be maintained)
- The just-added `get`/`set`/`copyModifiedTime` will be deprecated, so that they may be eventually removed.
- In a future version of sbt/io, once we remove the `modifiedTime` calls, we will add instead the versions that return `Try` (or `Either`), so that the different types can guide us into a migration. The zero-returning variants will be deprecated.

In conclusion, eventually we will have only three get/set/copy calls, based on `Try`.

cunei:
@dwijnand, since #82 is closed I re-enabled the pending modified files detection test; please review.

eed3si9n (dismissed):
LGTM. Thanks for your patience on this Toni!

eed3si9n (approved):

cunei:
@eed3si9n I added an additional warning removal commit and your approval went away; please approve again. Thanks!

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/337/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/3822/head-pull/470/head-pull/106/head-pull/192/head-pull/136//head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/337/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3822">pull/3822/head      </A> | sbt/sbt@94e36a14c80027a4e316505574d4c49fdcb14948
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/470">pull/470/head       </A> | sbt/zinc@67d387856b3a30c757af306e2672d98f8630459b
  io                | <A HREF="https://github.com/sbt/io/pull/106">pull/106/head       </A> | sbt/io@77d3ba2fd47aa24be7cff3666223253024dea7ff
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/192">pull/192/head       </A> | sbt/librarymanagement@72dd0064c3791ad4a6a2ba1904d8cf12322aede3
  util              | <A HREF="https://github.com/sbt/util/pull/136/">pull/136//head      </A> | 
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3822/head&zinc=pull/470/head&io=pull/106/head&librarymanagement=pull/192/head&util=pull/136//head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/338/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/3822/head-pull/470/head-pull/106/head-pull/192/head-pull/136/head-1.x)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/338/">has checked</A> the following projects against Scala 2.12,
tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/3822">pull/3822/head      </A> | sbt/sbt@94e36a14c80027a4e316505574d4c49fdcb14948
  zinc              | <A HREF="https://github.com/sbt/zinc/pull/470">pull/470/head       </A> | sbt/zinc@67d387856b3a30c757af306e2672d98f8630459b
  io                | <A HREF="https://github.com/sbt/io/pull/106">pull/106/head       </A> | sbt/io@77d3ba2fd47aa24be7cff3666223253024dea7ff
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/pull/192">pull/192/head       </A> | sbt/librarymanagement@72dd0064c3791ad4a6a2ba1904d8cf12322aede3
  util              | <A HREF="https://github.com/sbt/util/pull/136">pull/136/head       </A> | sbt/util@0a1bd5a3b25f95d25a5fef18844cb0acf61a425b
  website           | <A HREF="https://github.com/sbt/website/tree/1.x">1.x                 </A> | 

:white_check_mark: The result is: ***SUCCESS***
<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/3822/head&zinc=pull/470/head&io=pull/106/head&librarymanagement=pull/192/head&util=pull/136/head&website=1.x&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

cunei:
@eed3si9n The five pull requests are ready for merging now.

io/src/test/scala/sbt/internal/io/SourceModificationWatchSpec.scala
```diff
@@ -21,10 +21,9 @@ abstract class SourceModificationWatchSpec(
 
     IO.write(file, "foo")
 
-    // watchTest(parentDir) {
-    //   IO.write(file, "bar")
-    // }
-    pending // until fixed https://github.com/sbt/io/issues/82
+    watchTest(parentDir) {
+      IO.write(file, "bar")
+    }
```
dwijnand, line 26: This shouldn't've been enabled like this because it fails on HFS+. See https://github.com/sbt/io/pull/102/files#diff-32eb449ec95851dd791288244207a8f8R32.

PR #121, opened 2018-01-20
align exception from file-timestamp functions from different impl.

j-keck:
* The native implementations to get / set a file-timestamp throws
   a 'java.io.FileNotFoundException' where the Java based implementation
   throws a 'java.nio.file.NoSuchFileException' if the given file
   does not exists.

 * 'IO.getModifiedTimeOrZero' and 'IO.setModifiedTimeOrFalse' handles
   only 'java.io.FileNotFoundException'

 * Map 'NoSuchFileExeption' to 'FileNotFoundException' in the
   Java based implementation, so all implementations throws
   the same error.

fixes sbt/sbt/issues/3894

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Milli.scala
+++ b/io/src/main/scala/sbt/internal/io/Milli.scala
@@ -7,7 +7,7 @@ import java.io.File
 import java.util.Date
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
-import java.nio.file.Files
+import java.nio.file.{ Files, NoSuchFileException }
 import java.nio.file.{ Paths => JPaths }
 import java.nio.file.attribute.FileTime
 
@@ -319,11 +319,19 @@ private abstract class MilliMilliseconds extends Milli {
 
 private object JavaMilli extends MilliMilliseconds {
   def getModifiedTime(filePath: String): Long =
-    Files.getLastModifiedTime(JPaths.get(filePath)).toMillis
-  def setModifiedTime(filePath: String, mtime: Long): Unit = {
-    Files.setLastModifiedTime(JPaths.get(filePath), FileTime.fromMillis(mtime))
-    ()
-  }
+    mapNoSuchFileException(Files.getLastModifiedTime(JPaths.get(filePath)).toMillis)
+  def setModifiedTime(filePath: String, mtime: Long): Unit =
+    mapNoSuchFileException {
+      Files.setLastModifiedTime(JPaths.get(filePath), FileTime.fromMillis(mtime))
+      ()
+    }
+
+  private def mapNoSuchFileException[A](f: => A): A =
+    try {
+      f
+    } catch {
+      case e: NoSuchFileException => throw new FileNotFoundException(e.getFile).initCause(e)
+    }
 }
 
 object Milli {
--- a/io/src/test/scala/sbt/io/IOSpec.scala
+++ b/io/src/test/scala/sbt/io/IOSpec.scala
@@ -61,4 +61,8 @@ class IOSpec extends FlatSpec with Matchers {
     val u = IO.toURI(file("src") / "main" / "scala")
     assert(IO.toFile(u) == (file("src") / "main" / "scala"))
   }
+
+  "getModifiedTimeOrZero" should "return 0L if the file doesn't exists" in {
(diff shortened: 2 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=1.x&io=pull/121/head&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand (approved):
LGTM. Requesting review by @cunei for his thoughts.

dwijnand (changes requested):
Sorry @j-keck, do you mind rebasing (or cherry-picking) this commit onto the 1.1.x branch? That way hopefully we can merge it for sbt 1.1.1.

j-keck:
@dwijnand done, this pr is on top of 1.1.x

dwijnand (dismissed):
Thanks. Re-LGTM from me.

io/src/main/scala/sbt/internal/io/Milli.scala
```diff
@@ -319,11 +319,19 @@ private abstract class MilliMilliseconds extends Milli {
...
+    mapNoSuchFileException {
+      Files.setLastModifiedTime(JPaths.get(filePath), FileTime.fromMillis(mtime))
+      ()
+    }
+
+  private def mapNoSuchFileException[A](f: => A): A =
+    try {
+      f
+    } catch {
+      case e: NoSuchFileException => throw new FileNotFoundException(e.getFile)
```
cunei: An issue here is that the original stack trace is lost. It is not a big problem since `mapNoSuchFileException()` is only used once, but if it were used from more than one method, it would not be clear which code path led to the exception. Normally that can be avoided by chaining exceptions; however, it just so happens that `FileNotFoundException` does not have a `Throwable`-based constructor, so chaining is a bit more difficult than usual. Still, there is a workaround:
```
case e: NoSuchFileException => throw new FileNotFoundException(e.getFile) {
  initCause(e)
  override def toString() = {
    val msg = getLocalizedMessage()
    getClass().getSuperclass().getName() + (if (msg == null) "" else (": " + msg))
  }
}
```
This will be reported as:
```
java.io.FileNotFoundException: filename
        at ...
Caused by: java.nio.file.NoSuchFileException: filename
        at ...
```
Where the second exception identifies the original stacktrace. The additional `toString()` is needed in order to keep the exception identifying itself as a `FileNotFoundException` in the printout.

cunei (commented):
Thanks @j-keck, it looks good apart from a comment I left concerning stack traces.

io/src/main/scala/sbt/internal/io/Milli.scala
dwijnand: I think you can just call [`initCause`](https://docs.oracle.com/javase/8/docs/api/java/lang/Throwable.html#initCause-java.lang.Throwable-).

j-keck: thanks for your suggestions.
yes, `throw new FileNotFoundException(e.getFile).initCause(e)` preserves the stacktrace - thanks:
```
[error] java.io.FileNotFoundException: /nothing
[error] 	at sbt.internal.io.JavaMilli$.mapNoSuchFileException(Milli.scala:334)
[error] 	at sbt.internal.io.JavaMilli$.getModifiedTime(Milli.scala:322)
[error] 	at sbt.internal.io.Milli$.getModifiedTime(Milli.scala:362)
[error] 	at sbt.io.IO$.getModifiedTime(IO.scala:1143)
...
[error] Caused by: java.nio.file.NoSuchFileException: /nothing
[error] 	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:86)
[error] 	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
[error] 	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
...
```
should i make this change in a new commit, or amend the actual commit?

dwijnand: As you prefer.

j-keck:
done

dwijnand (approved):

dwijnand:
Thanks, @j-keck.

PR #138, opened 2018-03-29
Add MacOSXWatchService

eatkins:
This watch service is taken from my CloseWatch plugin, found at
https://github.com/swoval/swoval

It uses a native jni library, provided by apple-file-events, to receive
file events from the os. I more or less copied this file from the close
watch source code with some minor modifications given that it is now in
the sbt.io package.

### DIFF
--- a/io/src/main/scala/sbt/io/MacOSXWatchService.scala
+++ b/io/src/main/scala/sbt/io/MacOSXWatchService.scala
@@ -1,0 +1,134 @@
+package sbt.io
+
+import java.nio.file.StandardWatchEventKinds.{ ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY, OVERFLOW }
+import java.nio.file.{ Files, WatchEvent, WatchKey, Path => JPath, Paths => JPaths }
+import java.util.Collections
+import java.util.concurrent._
+import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger }
+import java.util.{ List => JList }
+
+import com.swoval.concurrent.ThreadFactory
+import com.swoval.files.apple.FileEventsApi.Consumer
+import com.swoval.files.apple.{ FileEvent, FileEventsApi, Flags }
+
+import scala.collection.mutable
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
+class MacOSXWatchService extends WatchService {
+  // The FsEvents api doesn't seem to report events at lower than 10 millisecond intervals.
+  private[this] val watchLatency: Duration = 10.milliseconds
+  private[this] val queueSize = 256
+
+  private[this] val executor =
+    Executors.newSingleThreadExecutor(new ThreadFactory("sbt.io.MacOSXWatchService"))
+  private[this] val dropEvent = new Consumer[String] {
+    override def accept(s: String): Unit = {}
+  }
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=pull/138/head&librarymanagement=1.1.x&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/MacOSXWatchService.scala
```diff
@@ -0,0 +1,123 @@
...
+
+import scala.collection.mutable
+import scala.collection.JavaConverters._
+import scala.concurrent.duration.Duration
+
+class MacOSXWatchService(watchLatency: Duration, val queueSize: Int = 256) extends WatchService {
+
+  private[this] val executor =
+    Executors.newSingleThreadExecutor(new ThreadFactory("sbt.io.MacOSXWatchService"))
+  private[this] val watcher: FileEventsApi = FileEventsApi.apply(onFileEvent, _ => {})
```
eed3si9n: This failed compilation on Travis for 2.10:

```scala
[error] /home/travis/build/sbt/io/io/src/main/scala/sbt/io/MacOSXWatchService.scala:21:66: missing arguments for method onFileEvent in class MacOSXWatchService;
[error] follow this method with `_' if you want to treat it as a partially applied function
[error]   private[this] val watcher: FileEventsApi = FileEventsApi.apply(onFileEvent, _ => {})
[error]                                                                  ^
```

io/src/main/scala/sbt/io/MacOSXWatchService.scala
```diff
@@ -0,0 +1,123 @@
...
+          registered += path -> key
+          val flags = new Flags.Create().setNoDefer().setFileEvents().value
+          watcher.createStream(path.toRealPath().toString, watchLatency.toMillis / 1000.0, flags)
+          key
+      }
+    }
+  }
+
+  def onFileEvent(fileEvent: FileEvent): Unit = {
+    executor.submit((() => {
```
eed3si9n:
```scala
[error] /home/travis/build/sbt/io/io/src/main/scala/sbt/io/MacOSXWatchService.scala:61:25: type mismatch;
[error]  found   : () => Unit
[error]  required: Runnable
[error]     executor.submit((() => {
[error]                         ^
```

eed3si9n (changes requested):
Travis CI failed for Scala 2.10. Probably need to make small adjustments to supply type annotation or sugar.

eatkins:
I made the file 2.10 compatible, but, out of curiosity, why do we build for 2.10 (and 2.11)? Isn't this module only used by sbt 1.x?

eed3si9n:
> I made the file 2.10 compatible, but, out of curiosity, why do we build for 2.10 (and 2.11)? Isn't this module only used by sbt 1.x?

We first split out IO as an aspirational general-purpose IO module while we were still in 0.13, and built other things on top of that. At some point we might drop 2.10.

eed3si9n (approved):

dwijnand (approved):
thanks @eatkins

PR #139, opened 2018-04-05
Expose JavaMilli to the whole sbt package

dwijnand:
Allows for reuse, such as in sbt/sbt#4067.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Milli.scala
+++ b/io/src/main/scala/sbt/internal/io/Milli.scala
@@ -1,15 +1,11 @@
 package sbt.internal.io
 
-import java.lang.UnsupportedOperationException
 import java.io.IOException
 import java.io.FileNotFoundException
 import java.io.File
 import java.util.Date
 import java.nio.ByteBuffer
 import java.nio.ByteOrder
-import java.nio.file.{ Files, NoSuchFileException }
-import java.nio.file.{ Paths => JPaths }
-import java.nio.file.attribute.FileTime
 
 import scala.reflect.{ ClassTag, classTag }
 import scala.collection.JavaConverters.mapAsJavaMapConverter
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=pull/139/head&librarymanagement=1.1.x&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/internal/io/Milli.scala
```diff
@@ -56,7 +56,7 @@ private abstract class StatInt(size: Int, mtimeOffset: Int, mtimensecOffset: Int
   def getModifiedTimeNative = (buffer.getInt(mtimeOffset), buffer.getInt(mtimensecOffset))
 }
 
-private abstract class Milli {
+private[sbt] abstract class Milli {
```
eed3si9n, line 56: I am of two minds on this one.
I thought we discussed this type of things previously and said we should keep internal code to each modules, and essentially copy-paste so they are less coupled.

cunei: I see nothing wrong in exposing internal code to `[sbt]`, it is still confined. Code duplication, however, is almost never a good idea.
This PR has my LGTM

cunei (dismissed):

io/src/main/scala/sbt/internal/io/Milli.scala
dwijnand: I don't mind making the whole thing public either.

dwijnand: by "whole thing" I mean `JavaMilli`, btw.

eed3si9n: Making it public is better I think. Let's move it to sbt.io package.

dwijnand:
done. please re-review @cunei and @eed3si9n.

eed3si9n (dismissed):
LGTM pending Travis

cunei (approved):
Looks ok!

PR #147, opened 2018-04-29
prepare Scala 2.13.0-M4

xuwei-k:
- use `Set#diff` instead of `Set#--` https://github.com/scala/scala/blob/5661290bfda0a08a60fb8df3ac60baafb55c0306/src/library/scala/collection/Set.scala#L154
- use `foldLeft` instead of `/:` https://github.com/scala/scala/blob/5661290bfda0a08a60fb8df3ac60baafb55c0306/src/library/scala/collection/IterableOnce.scala#L465


eed3si9n (approved):
Thanks!

dwijnand (approved):

PR #149, opened 2018-05-08
Add sleep to user input thread

eatkins:
I didn't realize that System.in.read is not interruptible, so it was
problematic* to change the downstream termination condition from:
-  System.in.available > 0 && (watched.terminateWatch(System.in.read()) || shouldTerminate)
+  watched.terminateWatch(System.in.read()) || shouldTerminate

Instead, we should revert to the old termination condition and add a
sleep so that we don't peg the cpu.

* This would manifest as the user having to hit enter twice to exit the
watch in some circumstances because the user input thread could not be
joined.


eed3si9n (approved):

eed3si9n:
ok. A new version of IO coming up.

PR #150, opened 2018-05-10
Fix file watching for unix

eatkins:
I inadvertently forgot to reset the watch key after reading the events.
This would cause the EventMonitor to miss watch events because the watch
service stopped posting events to the full key. I looked at the jdk
source code and it looks like each watch key can hold up to 512 events,
so I added a test to ensure that after adding 1000 files, watch events
were still triggered. The test failed before the change to EventMonitor.
I also realized that I reintroduced the bug fixed by @oneill in
881e835a56ef9f7df15075fd6abb62c122bae2af so I fixed that as well.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/EventMonitor.scala
+++ b/io/src/main/scala/sbt/internal/io/EventMonitor.scala
@@ -130,8 +130,11 @@ private[sbt] object EventMonitor {
       def getFilesForKey(key: WatchKey): collection.Seq[Path] = key match {
         case null => Nil
         case k =>
-          val allEvents = k.pollEvents.asScala
-            .map(e => k.watchable.asInstanceOf[Path].resolve(e.context.asInstanceOf[Path]))
+          val rawEvents = k.pollEvents
+          k.reset()
+          val allEvents = rawEvents.asScala.flatMap { e =>
+            Option(e.context).map(c => k.watchable.asInstanceOf[Path].resolve(c.asInstanceOf[Path]))
+          }
           logger.debug(s"Received events:\n${allEvents.mkString("\n")}")
           val (exist, notExist) = allEvents.partition(Files.exists(_))
           val (updatedDirectories, updatedFiles) = exist.partition(Files.isDirectory(_))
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=pull/150/head&librarymanagement=1.1.x&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (dismissed):
LGTM
Thanks for the tests too

eatkins:
Yeah, I was pretty disappointed that I didn't manually test this change fully under linux so I wanted to make sure that there was a regression test. The most annoying part is that there is very similar code in close watch that did the right thing but I didn't refer to it when I implemented the EventMonitor.

dwijnand (approved):

OlegYch:
i tried this on windows and it doesn't seem to make any difference (compared to sbt 1.1.5) - ~ works only once, and then all modifications are ignored until i relaunch it

eatkins:
@OlegYch How did you test this change? Did you build an entirely new version of sbt with the updated io dependency? I ask because I've tried making my project depend on a new version of io via library dependencies and that is not sufficient to get sbt to use the new io. At any rate, I'm downloading a windows iso to see if I can reproduce the bug on a windows vm.

OlegYch:
the added tests seems to fail too https://travis-ci.org/sbt/io/jobs/378830446 , though only on travis

OlegYch:
@eatkins no, i built io locally and then added
`libraryDependencies += "org.scala-sbt" %% "io" % "1.1.9-c3c8848d479bee7f0c9004c505e08631f2575a8a-SNAPSHOT"` to one of my projects `project/plugins.sbt`
confirmed it was in effect by doing `;reload plugins; show fullClasspath`

OlegYch:
btw any idea why this https://github.com/sbt/io/blob/1.1.x/build.sbt#L15 says  1.1.4 and not 1.1.8 which is the latest published version?

OlegYch:
@eatkins you're right simple libdeps override doesn't work, even if fullClasspath says otherwise
i rebuilt sbt and your solution works as intended on windows too, sorry for the noise
test failure on travis is a concern though

eatkins:
Agreed about travis. I am struggling to reproduce, but it's possible that it might be fixed by: https://github.com/sbt/io/pull/151/commits/5fc1e4677d97c77003d465ba9f79bbeb7f325805

eatkins:
Also, on the plus side, I now have a windows vm for testing which may be useful.

PR #148, opened 2018-05-08
fix compile error in Scala 2.13.0-M4

xuwei-k:
- https://github.com/scala/community-builds/issues/710
- https://scala-ci.typesafe.com/job/scala-2.13.x-integrate-community-build/1092/consoleFull

```
[sbt-io] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.12/project-builds/sbt-io-1aeb11a3d751ba13cb0d1d22d082306f3c24a209/io/src/main/scala/sbt/io/MacOSXWatchService.scala:61:32: overloaded method value flatMap with alternatives:
[sbt-io] [error]   [K2, V2](f: ((java.nio.file.Path, sbt.io.MacOSXWatchKey)) => scala.collection.IterableOnce[(K2, V2)])scala.collection.mutable.Map[K2,V2] <and>
[sbt-io] [error]   [B](f: ((java.nio.file.Path, sbt.io.MacOSXWatchKey)) => scala.collection.IterableOnce[B])scala.collection.mutable.Iterable[B]
[sbt-io] [error]  cannot be applied to (((java.nio.file.Path, sbt.io.MacOSXWatchKey)) => Option[(sbt.io.MacOSXWatchKey, scala.collection.mutable.Buffer[java.nio.file.WatchEvent[java.nio.file.Path]])])
[sbt-io] [error]       .synchronized(registered.flatMap {
[sbt-io] [error]                                ^
[sbt-io] [error] /home/jenkins/workspace/scala-2.13.x-integrate-community-build/target-0.9.12/project-builds/sbt-io-1aeb11a3d751ba13cb0d1d22d082306f3c24a209/io/src/main/scala/sbt/io/MacOSXWatchService.scala:61:32: overloaded method value flatMap with alternatives:
[sbt-io] [error]   [K2, V2](f: ((java.nio.file.Path, sbt.io.MacOSXWatchKey)) => scala.collection.IterableOnce[(K2, V2)])scala.collection.mutable.Map[K2,V2] <and>
[sbt-io] [error]   [B](f: ((java.nio.file.Path, sbt.io.MacOSXWatchKey)) => scala.collection.IterableOnce[B])scala.collection.mutable.Iterable[B]
[...]

PR #151, opened 2018-05-14
Prefer immutable.Seq & generic Iterable over generic Seq

dwijnand:
Alternative fix to #148.

/cc @xuwei-k

### DIFF
--- a/io/src/main/scala/sbt/internal/io/EventMonitor.scala
+++ b/io/src/main/scala/sbt/internal/io/EventMonitor.scala
@@ -7,7 +7,6 @@ import java.util.concurrent.atomic.{ AtomicBoolean, AtomicInteger }
 import sbt.io.WatchService
 
 import scala.annotation.tailrec
-import scala.collection.Iterable
 import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 
@@ -128,10 +127,10 @@ private[sbt] object EventMonitor {
         recentEvents = recentEvents.filterNot(_._2.isOverdue)
         getFilesForKey(s.service.poll(delay)).foreach(maybeTrigger)
       }
-      def getFilesForKey(key: WatchKey): Iterable[Path] = key match {
-        case null => Nil
+      def getFilesForKey(key: WatchKey): Vector[Path] = key match {
+        case null => Vector.empty
         case k =>
-          val allEvents = k.pollEvents.asScala
+          val allEvents = k.pollEvents.asScala.toVector
             .map(e => k.watchable.asInstanceOf[Path].resolve(e.context.asInstanceOf[Path]))
           logger.debug(s"Received events:\n${allEvents.mkString("\n")}")
           val (exist, notExist) = allEvents.partition(Files.exists(_))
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=pull/151/head&librarymanagement=1.1.x&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand:
@eatkins would you like to review this, by chance?

io/src/main/scala/sbt/internal/io/EventMonitor.scala
```diff
@@ -127,12 +127,12 @@ private[sbt] object EventMonitor {
         recentEvents = recentEvents.filterNot(_._2.isOverdue)
         getFilesForKey(s.service.poll(delay)).foreach(maybeTrigger)
       }
-      def getFilesForKey(key: WatchKey): collection.Seq[Path] = key match {
-        case null => Nil
+      def getFilesForKey(key: WatchKey): Vector[Path] = key match {
+        case null => Vector.empty
         case k =>
-          val rawEvents = k.pollEvents
+          val rawEvents = k.pollEvents.asScala.toVector
```
eatkins: I was thinking about this line and I think that it would make sense to change it to:
```scala
val rawEvents = k.synchronized {
  val events = k.pollEvents.asScala.toVector
  k.reset()
  events
}
```
The reason is that k.reset() clears out the events for the key in the default watch service. This makes it possible to miss an event if it added between the call to pollEvents and reset (I think). That is why I put the k.reset() immediately after the call to k.pollEvents and deferred any processing of the events until after the call to reset(). I looked again at the open jdk code and all of its public methods use `synchronized(this)` so synchronization on the key should hopefully prevent us from ever missing any events.

io/src/main/scala/sbt/internal/io/EventMonitor.scala
```diff
@@ -127,12 +127,12 @@ private[sbt] object EventMonitor {
         recentEvents = recentEvents.filterNot(_._2.isOverdue)
         getFilesForKey(s.service.poll(delay)).foreach(maybeTrigger)
       }
-      def getFilesForKey(key: WatchKey): collection.Seq[Path] = key match {
-        case null => Nil
+      def getFilesForKey(key: WatchKey): Vector[Path] = key match {
```
eatkins, line 130: As a rule, I try to leave apis as generic as possible, so I'd probably use `IndexedSeq` over `Vector`, but I don't have strong feelings about this. This isn't a public api anyway.

eatkins (approved):
LGTM

io/src/main/scala/sbt/internal/io/EventMonitor.scala
dwijnand: Eugene argued for Vector over Iterable, and I didn't have strong feelings about it either.

eed3si9n: My general preference is concrete datatypes instead of mixing and matching - https://www.scala-sbt.org/1.x/docs/Coding-Guideline.html#Use+specific+datatypes+%28%2C+%2C+or+%29%2C+rather+than
We might still opt to keep things abstract for performance reasons, but I think that should an exception not the rule.

PR #134, opened 2018-02-26
WinMilli: enable read/read/delete/rename sharing

cunei:
In Windows, reading or writing file attributes requires opening
the file via CreateFile(). If only FILE_SHARE_READ is specified
when reading the timestamp, the CreateFile() call will fail
when the file happens to be already open for writing. This
pull request specifies that the file that is being opened has
shared read and write access; the shared delete mode (which also
implies rename) is also allowed, and in case the subsequent
timestamp get/set fails (if the file disappears) the usual
exception-to-zero code path will be taken.

See sbt/sbt#3972

PR #160, opened 2018-05-28
Appveyor

eatkins:


eatkins:
The appveyor build is working for my personal account: https://ci.appveyor.com/project/eatkins/io. Someone with appropriate permissions would have to turn on appveyor for the io project.

dwijnand:
@cunei Could you please review the Milli changes in this PR, please?

dwijnand (dismissed):
I'm happy to land this change, so LGTM.

io/src/main/scala/sbt/internal/io/Milli.scala
```diff
@@ -281,12 +290,16 @@ private object WinMilli extends MilliNative[FILETIME] {
                            null)
     if (hFile == INVALID_HANDLE_VALUE) {
       val err = GetLastError()
-      if (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND)
+      if (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND || err == ERROR_ACCESS_DENIED)
         throw new FileNotFoundException("Not found: " + lpFileName)
-      else
+      else if (err == ERROR_SHARING_VIOLATION && attempt < MAX_GET_HANDLE_ATTEMPTS) {
+        Thread.sleep(1)
+        getHandle(lpFileName, dwDesiredAccess, dwShareMode, attempt + 1)
```
cunei: The `ERROR_SHARING_VIOLATION` should (in theory) already be fixed by this pull request: https://github.com/sbt/io/pull/134
That PR was merged in 1.x (on March 7th) but it is not in 1.1.x. The 1.1.x branch is still very much active at this point, so it may be worth backporting it, rather than using workarounds. /cc @dwijnand

cunei: In https://github.com/sbt/sbt/issues/3972#issuecomment-384634235, a user reports that an `ERROR_SHARING_VIOLATION` still appears in a 1.x snapshot when running under the Windows Subsystem for Linux (with #134 applied); conversely in https://github.com/sbt/sbt/issues/3972#issuecomment-369457527 the patch seems to have resolved the issue.
Access permissions are weird in Windows, and I was unable to reproduce it post-patch. In the context of this pull request, I would start by backporting #134 to 1.1.x, and use the workaround proposed by @eatkins only if that is still a problem in Appveyor afterwards. That would be my suggestion, at least.

io/src/main/scala/sbt/internal/io/Milli.scala
```diff
@@ -281,12 +290,16 @@ private object WinMilli extends MilliNative[FILETIME] {
                            null)
     if (hFile == INVALID_HANDLE_VALUE) {
       val err = GetLastError()
-      if (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND)
+      if (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND || err == ERROR_ACCESS_DENIED)
```
cunei: Good catch. An alternative could be throwing `java.nio.file.AccessDeniedException` for `ERROR_ACCESS_DENIED`, but `java.io.FileNotFoundException` was the usual response before nio, so I'm ok with that. I'd just use a different string for the error message in that case, for clarity, like:
```
if (err == ERROR_FILE_NOT_FOUND || err == ERROR_PATH_NOT_FOUND)
  throw new FileNotFoundException("Not found: " + lpFileName)
else if (err == ERROR_ACCESS_DENIED)
  throw new FileNotFoundException("Access denied: " + lpFileName)
else
```

cunei (changes requested):

eatkins:
@cunei thanks for the suggestions. I'll incorporate them soon and re-push.

eatkins:
I closed this in favor of #173, which is compared against 1.2.x instead of 1.1.x. @cunei was right that my transient failures on appveyor went away so I was able to get rid of the polling commit.

PR #177, opened 2018-08-29
File watch overhaul

eatkins:
This is a large pull request that is the culmination of a number of month's work trying to improve the overall io and watch performance of sbt. For the most part, I've broken the commits up so that each one more or less stands on its own. I am doing it as one large pull request just because the overhead of making 20 partially dependent commits feels to high.

This commit uses apis that I developed in the swoval [file-tree-views](https://github.com/swoval/swoval/blob/master/files/README.md) library that improve the performance of traversing the file system tree and also provide a dynamically updating in-memory cache of the file system.

As I describe in the file-tree-views readme, it is surprisingly expensive to descend the file system tree because all of the built in jvm apis require a stat of each file to determine if its a directory or a regular file. In the swoval library, I provide jni native libraries that take advantage of extra information in the system call for listing directories to traverse the tree without stat-ing any of the files. This can reduce the run time of traversal of a directory by 3-5x. It was straightforward to update the PathFinder implementations to use the swoval implementations. Because there is some risk that there is a bad bug in the library, I still allow users to fallback to the built-in java implementations via a system property (I have never seen any unexpected behavior, but I can't rule it out in every environment).

Secondly, I add support to cache a portion of the file system. The motivation is so that we can reduce the latency between the beginning of the compile task to when we actually start compilation. Zinc needs to compare file hashes so right now, sbt has to list all the files and compute the hash. The hash is fast, but the entire process is highly io bound. I seem to get roughly 10K iops on my mac with an ssd so it is perhaps unsurprising that before starting compilation I see sbt doing about 700ms of io before compilation starts in a project with 5000 files. By instead using an in-memory cache of the file system that precomputes the hashes, I've seen that 700ms drop drop to about 100-150. As a bonus, since we are already monitoring the file system to maintain the cache, we can re-use the cache in continuous builds by adding a listener to the cache's observable.

Finally, I make some improvements to the watch api for sbt. I refactor the EventMonitor so that it returns more information to sbt. I use this to make continuous builds testable by allowing the user to provide arbitrary termination conditions which do not depend on reading from System.in. In addition, I am able to implement automatic project reloading when the meta build files change. This is a feature that mill advertises as an improvement over sbt: http://www.lihaoyi.com/post/MillBetterScalaBuilds.html. 

I have worked as hard as I could to try and preserve all existing behavior for users. In a number of places, I provide multiple implementations that are user configurable via a system property or sbt setting. In an ideal world, the majority of sbt users notice nothing after the change except that sbt feels snappier than before.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,7 @@ val io = (project in file("io"))
     name := "IO",
     libraryDependencies ++= {
       Vector(scalaCompiler.value % Test, scalaCheck % Test, scalatest.value % Test)
-    } ++ Vector(appleFileEvents),
+    } ++ Vector(swovalFiles),
     libraryDependencies ++= Seq(jna, jnaPlatform),
     sourceManaged in (Compile, generateContrabands) := baseDirectory.value / "src" / "main" / "contraband-scala",
     initialCommands in console += "\nimport sbt.io._, syntax._",
--- a/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
+++ b/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
@@ -1,0 +1,47 @@
+package sbt.internal.io
+
+import java.nio.file.{ NoSuchFileException, NotDirectoryException, Path }
+
+import com.swoval.files.{ TypedPath => STypedPath }
+import com.swoval.files.FileTreeViews
+import com.swoval.functional.Filters
+import sbt.io.{ FileTreeView, TypedPath }
+
+import scala.collection.JavaConverters._
+import SwovalConverters._
+
+import scala.util.Try
+
+private[sbt] object DefaultFileTreeView extends FileTreeView {
+  private[this] val fileTreeView =
+    if ("nio" == System.getProperty("sbt.pathfinder.implementation"))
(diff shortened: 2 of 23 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.2.x&zinc=1.2.x&io=pull/177/head&librarymanagement=1.2.x&util=1.2.x&website=1.2.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
@eatkins Thanks for this contribution.

I'd be happy to adopt your new file watching improvements, but in terms of the branch, I'd vote to put this into 1.x (future 1.3.0) per discussions in https://discuss.lightbend.com/t/patching-strategy/1924. Your feedback there would be welcome on this topic.

eatkins:
@eed3si9n should I also re-do the sbt PR to be compared against 1.x?

eed3si9n:
> should I also re-do the sbt PR to be compared against 1.x?

Yes. We could've also used GitHub UI to retarget to 1.x.

PR #178, opened 2018-08-29
File watch overhaul

eatkins:
This is a large pull request that is the culmination of a number of month's work trying to improve the overall io and watch performance of sbt. For the most part, I've broken the commits up so that each one more or less stands on its own. I am doing it as one large pull request just because the overhead of making 20 partially dependent commits feels to high.

This commit uses apis that I developed in the swoval file-tree-views library that improve the performance of traversing the file system tree and also provide a dynamically updating in-memory cache of the file system.

As I describe in the file-tree-views readme, it is surprisingly expensive to descend the file system tree because all of the built in jvm apis require a stat of each file to determine if its a directory or a regular file. In the swoval library, I provide jni native libraries that take advantage of extra information in the system call for listing directories to traverse the tree without stat-ing any of the files. This can reduce the run time of traversal of a directory by 3-5x. It was straightforward to update the PathFinder implementations to use the swoval implementations. Because there is some risk that there is a bad bug in the library, I still allow users to fallback to the built-in java implementations via a system property (I have never seen any unexpected behavior, but I can't rule it out in every environment).

Secondly, I add support to cache a portion of the file system. The motivation is so that we can reduce the latency between the beginning of the compile task to when we actually start compilation. Zinc needs to compare file hashes so right now, sbt has to list all the files and compute the hash. The hash is fast, but the entire process is highly io bound. I seem to get roughly 10K iops on my mac with an ssd so it is perhaps unsurprising that before starting compilation I see sbt doing about 700ms of io before compilation starts in a project with 5000 files. By instead using an in-memory cache of the file system that precomputes the hashes, I've seen that 700ms drop drop to about 100-150. As a bonus, since we are already monitoring the file system to maintain the cache, we can re-use the cache in continuous builds by adding a listener to the cache's observable.

Finally, I make some improvements to the watch api for sbt. I refactor the EventMonitor so that it returns more information to sbt. I use this to make continuous builds testable by allowing the user to provide arbitrary termination conditions which do not depend on reading from System.in. In addition, I am able to implement automatic project reloading when the meta build files change. This is a feature that mill advertises as an improvement over sbt: http://www.lihaoyi.com/post/MillBetterScalaBuilds.html.

I have worked as hard as I could to try and preserve all existing behavior for users. In a number of places, I provide multiple implementations that are user configurable via a system property or sbt setting. In an ideal world, the majority of sbt users notice nothing after the change except that sbt feels snappier than before.

There also was an issue that watching a nested symlinked directory did not work on osx. This was because osx file watchers are all recursive, so I'd added logic to the swoval file events api that pruned any watches for any watches of a descendant of a watched directory. When there are no symlinks, this works great and lowers the resource cost of the watch, but in the presence of a symlinked directory descendant, we'd lose the updates. I fixed this upstream in swoval and I even enhanced it so that updates of symlinked files are detected by the watch service. This enhancement is only available using the swoval file repository and will not be available using the WatchServiceBackedObservable implementation of the FileEventMonitor.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
+++ b/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
@@ -1,0 +1,53 @@
+package sbt.internal.io
+
+import java.nio.file.{ NoSuchFileException, NotDirectoryException, Path }
+
+import com.swoval.files.{ TypedPath => STypedPath }
+import com.swoval.files.FileTreeViews
+import com.swoval.functional.Filters
+import sbt.io.{ FileTreeView, TypedPath }
+
+import scala.collection.JavaConverters._
+import SwovalConverters._
+
+import scala.util.Try
+
+private[sbt] object DefaultFileTreeView extends FileTreeView {
+  private[this] val fileTreeView =
+    if ("nio" == sys.props.getOrElse("sbt.pathfinder", ""))
+      FileTreeViews.getNio(true)
+    else
+      FileTreeViews.getDefault(true)
+
+  /**
+   * List the contents of the current directory
+   *
+   * @param path      the path to list
+   * @param maxDepth  controls the depth of children of the path to include in the results. When
+   *                  maxDepth is -1, [[list]] should only return the TypedPath for this directory.
+   *                  For non-negative values, [[list]] should return only entries whose relativized
+   *                  path has {{{maxDepth - 1}}} elements. For example, when maxDepth is zero, all of
+   *                  the children of the path should be included in the result, but none of the
+   *                  children of any of the subdirectories should be incldued.
+   *
+   * @param filter    only return files accepted by the filter
+   * @return a sequence of [[TypedPath]].
+   */
+  override def list(path: Path, maxDepth: Int, filter: TypedPath => Boolean): Seq[TypedPath] = {
+    try {
+      fileTreeView
+        .list(path, maxDepth, Filters.AllPass)
+        .asScala
+        .toSeq
+        .flatMap((_: STypedPath).asSbt match {
+          case tp if filter(tp) => Some(tp)
+          case _                => None
+        })
+    } catch {
+      case _: NotDirectoryException | _: NoSuchFileException =>
+        Try(Seq(TypedPath(path))).getOrElse(Nil)
+    }
+  }
+
+  override def close(): Unit = ()
+}
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -18,5 +18,5 @@ object Dependencies {
   }
   val jna = "net.java.dev.jna" % "jna" % "4.5.0"
   val jnaPlatform = "net.java.dev.jna" % "jna-platform" % "4.5.0"
-  val appleFileEvents = "com.swoval" % "apple-file-events" % "1.3.2"
+  val swovalFiles = "com.swoval" % "file-tree-views" % "2.0.0"
 }
--- a/build.sbt
+++ b/build.sbt
@@ -64,6 +64,16 @@ val io = (project in file("io"))
 
       // protected[this]
       exclude[DirectMissingMethodProblem]("sbt.io.CopyOptions.copy*"),
+
+      // private class
+      exclude[MissingClassProblem]("sbt.io.Event"),
+      exclude[MissingClassProblem]("sbt.io.Event$"),
+      exclude[MissingClassProblem]("sbt.io.MacOSXWatchKey"),
+
+      // private internal classes whose functionality has been replaced
+      exclude[MissingClassProblem]("sbt.internal.io.EventMonitor$*"),
+      exclude[DirectMissingMethodProblem]("sbt.internal.io.EventMonitor.legacy"),
+      exclude[DirectMissingMethodProblem]("sbt.internal.io.EventMonitor.applyImpl")
     ),
     BuildInfoPlugin.buildInfoDefaultSettings, // avoids BuildInfo generated in Compile scope
     addBuildInfoToConfig(Test),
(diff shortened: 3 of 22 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=1.x&io=pull/178/head&librarymanagement=1.x&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
```diff
@@ -0,0 +1,47 @@
...
+import sbt.io.{ FileTreeView, TypedPath }
+
+import scala.collection.JavaConverters._
+import SwovalConverters._
+
+import scala.util.Try
+
+private[sbt] object DefaultFileTreeView extends FileTreeView {
+  private[this] val fileTreeView =
+    if ("nio" == System.getProperty("sbt.pathfinder.implementation"))
```
eed3si9n: `sys.props.get(...)` might be better. Also `implementation` in the name feels a bit awkward. I think we tend to call it something like `sbt.pathfinder` with "implementation" implied.

io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
```diff
@@ -0,0 +1,47 @@
...
+import com.swoval.functional.Filters
+import sbt.io.{ FileTreeView, TypedPath }
+
+import scala.collection.JavaConverters._
+import SwovalConverters._
+
+import scala.util.Try
+
+private[sbt] object DefaultFileTreeView extends FileTreeView {
+  private[this] val fileTreeView =
```
eed3si9n, line 16: Not sure how long it takes to initialize fileTreeView, but it might be better to use lazy val here.

eed3si9n (commented):
overall LGTM, I think.
left some minor comments.

io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
eatkins: I think it takes O(2-3ms) to load the swoval jni library if it hasn't been previously loaded. The actual initialization is near instantaneous otherwise.

io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
eatkins: Good point. I implemented the swoval library in java and got in the habit of using System.getProperty, so I'll switch to `sys.props.get` and change the name.

eatkins:
@eed3si9n Thanks for looking over the PR. I know that it's a lot.

project/Dependencies.scala
```diff
@@ -18,5 +18,5 @@ object Dependencies {
   }
   val jna = "net.java.dev.jna" % "jna" % "4.5.0"
   val jnaPlatform = "net.java.dev.jna" % "jna-platform" % "4.5.0"
-  val appleFileEvents = "com.swoval" % "apple-file-events" % "1.3.2"
+  val swovalFiles = "com.swoval" % "file-tree-views" % "2.0.0-RC2"
```
eed3si9n: It says RC2. Is it waiting on something?

eatkins: I have been reserving the right to change the public api until this PR is good to go. Moving forward, I'm going to enforce binary compatibility of the library in the 2.x series. As soon as we're ready to merge, I will release a 2.0.0 version and update the relevant commit in this PR.

eed3si9n: On develop branch I've added -Xfatal-warnings, so could you rebase against it? After that, I am good to go with this PR.

eatkins: Sure. I published swoval v2.0.0, rebased against origin/develop and updated Dependencies.scala.

eed3si9n (approved):

eed3si9n:
I should've run Dbuild before merging.

```
[sbt212] [error] /home/jenkinssbt/workspace/sbt-validator/target-0.9.14/project-builds/sbt212-0dcf5fdc6c1887745e8427a0496959076954d686/main/src/main/scala/sbt/TemplateCommandUtil.scala:66:12: reference to Logger is ambiguous;
[sbt212] [error] it is imported twice in the same scope by
[sbt212] [error] import sbt.util._
[sbt212] [error] and import sbt.io._
[sbt212] [error]       log: Logger
[sbt212] [error]            ^
[sbt212] [error] /home/jenkinssbt/workspace/sbt-validator/target-0.9.14/project-builds/sbt212-0dcf5fdc6c1887745e8427a0496959076954d686/main/src/main/scala/sbt/TemplateCommandUtil.scala:106:12: reference to Logger is ambiguous;
[sbt212] [error] it is imported twice in the same scope by
[sbt212] [error] import sbt.util._
[sbt212] [error] and import sbt.io._
[sbt212] [error]       log: Logger
[sbt212] [error]            ^
[sbt212] [error] /home/jenkinssbt/workspace/sbt-validator/target-0.9.14/project-builds/sbt212-0dcf5fdc6c1887745e8427a0496959076954d686/main/src/main/scala/sbt/TemplateCommandUtil.scala:134:12: reference to Logger is ambiguous;
[sbt212] [error] it is imported twice in the same scope by
[sbt212] [error] import sbt.util._
[sbt212] [error] and import sbt.io._
[sbt212] [error]       log: Logger
[sbt212] [error]            ^
[sbt212] [info] No documentation generated with unsuccessful compiler run
[sbt212] [error] three errors found
[sbt212] [error] (mainProj / Compile / compileIncremental) Compilation failed
[sbt212] [error] (mainProj / Compile / doc) Scaladoc generation failed
```

PR #175, opened 2018-08-02
Update relativize method for java 9

eatkins:
I noticed the io spec was failing on my mac on jdk9 (but passing on
travis, which was running jdk 8). The relativize method arguably didn't
work correctly on pre jdk 9 jres (at least on osx). For example, I ran
the following in the console:

scala> import java.nio.file._

scala> val base = Paths.get("/foo/bar/..")
val base = Paths.get("/foo/bar/..")
base: java.nio.file.Path = /foo/bar/..

scala> val file = Paths.get("/foo/buzz")
val file = Paths.get("/foo/buzz")
file: java.nio.file.Path = /foo/buzz

scala> val relative = base.relativize(file)
val relative = base.relativize(file)
                            ^
relative: java.nio.file.Path = ../../buzz

scala> base.resolve(relative).normalize
base.resolve(relative).normalize
                            ^
res5: java.nio.file.Path = /buzz

This result is certainly not what I would expect. The fix is very easy,
just always normalize the paths (which is not very expensive, especially
compared to any file system operation).

Issue #85, opened 2017-11-26
IOSpec and PollingWatchServiceSpec fail with Java 9


xuwei-k:
- https://travis-ci.org/xuwei-k/io/jobs/307362942
- https://github.com/xuwei-k/io/commit/f770321f1c1bcd838e420dd49e3e823637f53eaf

```
[info] IOSpec:
[info] - should relativize *** FAILED ***
[info]   Some("meh.file") was not equal to Some("../../meh.file") (IOSpec.scala:24)
```

```
[info] WatchService.close
[info] - should not throw if called multiple times
[info] + NameFilter.Exact filter matches valid string: OK, passed 100 tests.
[info] + NameFilter.Glob filter matches valid string if no *s: OK, passed 100 tests.
[info] + NameFilter.Exact filter matches provided string: OK, passed 100 tests.
[info] + NameFilter.All pass accepts everything: OK, passed 100 tests.
[info] + Copy.same contents: OK, passed 100 tests.
[info] + NameFilter.Glob filter matches provided string if no *s: OK, passed 100 tests.
[info] ! IO.IO.classLocationFile able to determine containing directories: Falsified after 1 passed tests.
[info] > ARG_0: class java.lang.Integer
[info] ScalaCheck
[info] Failed: Total 15, Failed 1, Errors 0, Passed 14
```

```
[info] Run completed in 59 seconds, 300 milliseconds.
[info] Total number of tests run: 63
[info] Suites: completed 8, aborted 0
[info] Tests: succeeded 62, failed 1, canceled 0, ignored 0, pending 0
[info] *** 1 TEST FAILED ***
[error] Failed: Total 78, Failed 2, Errors 0, Passed 76
[error] Failed tests:
[error] 	sbt.io.IOSpecification
[error] 	sbt.io.IOSpec
[error] (io/test:test) sbt.TestsFailedException: Tests unsuccessful
[error] Total time: 68 s, completed Nov 26, 2017, 4:39:18 AM
```

xuwei-k:
`IO.classLocationFile(cl: Class[_]): File` does not work with Java 9.

https://github.com/sbt/io/blob/b5a4b6f56e95945274d4d67680a73f1ffcb28a07/io/src/main/scala/sbt/io/IO.scala#L66-L74

Java 8

```
Welcome to Scala 2.12.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_151).
Type in expressions for evaluation. Or try :help.

[...]

PR #183, opened 2018-09-27
Fix IO.relativize

### LINKED ISSUE #182: IO.relativize is broken when base contains ".."
The inconsisnt behavior in JDK 11 reported in #85 exposed that `IO.relativize` is currently incorrect.

## steps

Given that the base directory is `/tmp/io-relativize/inside-dir/..` and that the target to be relativized is `/tmp/io-relativize/meh.file`, IOSpec states

```scala
IO.relativize(relativeRootDir, nestedFile).map(file) shouldBe Some(file("../../meh.file"))
```

## problem

`Some(file("../../meh.file"))` is incorrect.

## expectation

`Some("meh.file")`

## notes

```scala
    val basePath = (if (base.isAbsolute) base else base.getCanonicalFile).toPath
    val filePath = (if (file.isAbsolute) file else file.getCanonicalFile).toPath
    if ((filePath startsWith basePath) || (filePath.normalize() startsWith basePath.normalize())) {
      val relativePath = catching(classOf[IllegalArgumentException]) opt (basePath relativize filePath)
      relativePath map (_.toString)
    } else None
  }
```

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#isAbsolute:

> On UNIX systems, a pathname is absolute if its prefix is "/".

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#getCanonicalPath():

> This typically involves removing redundant names such as "." and ".." from the pathname, resolving symbolic links (on UNIX platforms)

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html#normalize()

> Returns a path that is this path with redundant name elements eliminated.

See also https://github.com/sbt/io/pull/175

eed3si9n:
Fixes #182 
See also https://github.com/sbt/io/pull/175 

The inconsisnt behavior in JDK 11 reported in #85 exposed that `IO.relativize` is currently incorrect.

The IOSpec states

```scala
IO.relativize(relativeRootDir, nestedFile).map(file) shouldBe Some(file("../../meh.file"))
```

however, given that the base directory is `/tmp/io-relativize/inside-dir/..` and that the target to be relativized is `/tmp/io-relativize/meh.file`, the correct answer should be `"meh.file"`, not `"../../meh.file"`.

This fixes the problem by avoiding the call to `getCanonicalFile`, which has the side effect of resolving symbolic links.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -18,7 +18,7 @@ import java.io.{
 import java.io.{ ObjectInputStream, ObjectStreamClass, FileNotFoundException }
 import java.net.{ URI, URISyntaxException, URL }
 import java.nio.charset.Charset
-import java.nio.file.FileSystems
+import java.nio.file.{ FileSystems, Path => NioPath }
 import java.nio.file.attribute.PosixFilePermissions
 import java.util.Properties
 import java.util.jar.{ Attributes, JarEntry, JarOutputStream, Manifest }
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/183/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #184, opened 2018-09-28
Fixes IO.relativize's handling of symbolic links

### LINKED ISSUE #182: IO.relativize is broken when base contains ".."

eed3si9n:
Backport of https://github.com/sbt/io/pull/183

Fixes #182 
See also https://github.com/sbt/io/pull/175 

The inconsistent behavior in JDK 11 reported in #85 exposed that `IO.relativize` is currently incorrect.

The IOSpec states

```scala
IO.relativize(relativeRootDir, nestedFile).map(file) shouldBe Some(file("../../meh.file"))
```

however, given that the base directory is `/tmp/io-relativize/inside-dir/..` and that the target to be relativized is `/tmp/io-relativize/meh.file`, the correct answer should be `"meh.file"`, not `"../../meh.file"`.

This fixes the problem by avoiding the call to `getCanonicalFile`, which has the side effect of resolving symbolic links.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -18,7 +18,7 @@ import java.io.{
 import java.io.{ ObjectInputStream, ObjectStreamClass, FileNotFoundException }
 import java.net.{ URI, URISyntaxException, URL }
 import java.nio.charset.Charset
-import java.nio.file.FileSystems
+import java.nio.file.{ FileSystems, Path => NioPath }
 import java.nio.file.attribute.PosixFilePermissions
 import java.util.Properties
 import java.util.jar.{ Attributes, JarEntry, JarOutputStream, Manifest }
@@ -608,8 +608,22 @@ object IO {
    * If `file` or `base` are not absolute, they are first resolved against the current working directory.
    */
   def relativize(base: File, file: File): Option[String] = {
-    val basePath = (if (base.isAbsolute) base else base.getCanonicalFile).toPath.normalize()
-    val filePath = (if (file.isAbsolute) file else file.getCanonicalFile).toPath.normalize()
+    // "On UNIX systems, a pathname is absolute if its prefix is "/"."
+    // https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#isAbsolute
+    // "This typically involves removing redundant names such as "." and ".." from the pathname, resolving symbolic links (on UNIX platforms)"
+    // https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#getCanonicalPath()
+    // We do not want to use getCanonicalPath because if we resolve the symbolic link, that could change
+    // the outcome of copyDirectory's target structure.
+    // Path#normailize is able to expand ".." without expanding the symlink.
+    // https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html#normalize()
+    // "Returns a path that is this path with redundant name elements eliminated."
+    def toAbsolutePath(x: File): NioPath = {
+      val p = x.toPath
+      if (!p.isAbsolute) p.toAbsolutePath
+      else p
+    }
+    val basePath = toAbsolutePath(base).normalize
+    val filePath = toAbsolutePath(file).normalize
     if (filePath startsWith basePath) {
       val relativePath = catching(classOf[IllegalArgumentException]) opt (basePath relativize filePath)
       relativePath map (_.toString)
--- a/io/src/test/scala/sbt/io/CopyDirectorySpec.scala
+++ b/io/src/test/scala/sbt/io/CopyDirectorySpec.scala
@@ -1,10 +1,10 @@
 package sbt.io
 
 import java.nio.file._
-import org.scalatest.{ FlatSpec, Matchers }
+import org.scalatest.{ FlatSpec, DiagrammedAssertions }
 import sbt.io.syntax._
 
-class CopyDirectorySpec extends FlatSpec with Matchers {
+class CopyDirectorySpec extends FlatSpec with DiagrammedAssertions {
   it should "copy symlinks" in IO.withTemporaryDirectory { dir =>
     // Given:
     // src/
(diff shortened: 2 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.2.x&zinc=1.2.x&io=pull/184/head&librarymanagement=1.2.x&util=1.2.x&website=1.2.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -608,8 +608,22 @@ object IO {
...
    */
   def relativize(base: File, file: File): Option[String] = {
-    val basePath = (if (base.isAbsolute) base else base.getCanonicalFile).toPath.normalize()
-    val filePath = (if (file.isAbsolute) file else file.getCanonicalFile).toPath.normalize()
+    // "On UNIX systems, a pathname is absolute if its prefix is "/"."
+    // https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#isAbsolute
+    // "This typically involves removing redundant names such as "." and ".." from the pathname, resolving symbolic links (on UNIX platforms)"
+    // https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#getCanonicalPath()
+    // We do not want to use getCanonicalPath because if we resolve the symbolic link, that could change
+    // the outcome of copyDirectory's target structure.
```
eatkins, line 616: Not only that, symlink resolution is quite expensive (I haven't benchmarked getCanonicalFile, but Path.toRealPath is even slower than Files.readAttributes).

io/src/test/scala/sbt/io/CopyDirectorySpec.scala
```diff
@@ -25,6 +25,6 @@ class CopyDirectorySpec extends FlatSpec with Matchers {
     IO.copyDirectory(dir / "src", dir / "dst")
 
     // Then: dst/lib/a.txt should have been created and have the correct contents
-    IO.read(dir / "dst" / "lib" / "a.txt") shouldBe "this is the file contents"
+    assert(IO.read(dir / "dst" / "lib" / "a.txt") == "this is the file contents")
```
eatkins, line 28: What's the motivation for moving away from the scalatest dsl to the more traditional assertion syntax?

io/src/test/scala/sbt/io/IOSpec.scala
```diff
@@ -15,14 +16,17 @@ class IOSpec extends FlatSpec with Matchers {
...
     val nestedFile = Files.createFile(rootDir resolve "meh.file").toFile
     val nestedDir = Files.createDirectory(rootDir resolve "inside-dir").toFile
 
     val relativeRootDir = new File(nestedDir, "..")
-
-    IO.relativize(rootDir.toFile, nestedFile).map(file) shouldBe Some(file("meh.file"))
-    IO.relativize(relativeRootDir, nestedFile).map(file) shouldBe Some(file("meh.file"))
+    assert(
+      IO.relativize(rootDir.toFile, nestedFile)
+        .map(normalizeForWindows) === Option("meh.file"))
```
eatkins, line 26: Is the change from `Some` to `Option` to fix type inference?

eatkins (approved):
LGTM.

io/src/test/scala/sbt/io/CopyDirectorySpec.scala
eed3si9n: ScalaTest has always supported `assert(...)` (http://www.scalatest.org/user_guide/using_assertions), but my overall preference has lately been moving away from BDD style to simpler unit tests, preferably with DiagrammedAssertions or [Expecty](http://eed3si9n.com/power-assert-with-expecty). I think they are easier to read/write/maintain.

io/src/test/scala/sbt/io/IOSpec.scala
eed3si9n: I am not sure if `Option(...)` change was necessary for ScalaTest's `===`, but many `Eq` typeclasses prefer the types to line up exactly on lhs and rhs.

io/src/test/scala/sbt/io/CopyDirectorySpec.scala
eatkins: Makes sense. I was just curious if this was a settled style decision moving forward. I don't have a horse in the race.

PR #186, opened 2018-09-30
Restore IO.classfileLocation behavior

eatkins:
The scripted tests fail for sbt when run against io/develop. This is
because zinc expects that IO.classfileLocation returns the full path of
a class file for any class that is in a directory, not the base path of
the class file directory.


eatkins:
Actually, I think this is wrong and it's zinc that should be updated instead of IO.

PR #187, opened 2018-09-30
Fix flaky test

eatkins:
This test occasionally fails on travis because the expected last
modified times don't match. This is likely because sometimes the
callback will fire before the file's last modified time has been
updated. To fix this, ensure that we only trigger the latch when the
file's last modified time has been updated to the expected value.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/187/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Thanks!

PR #196, opened 2018-10-08
Io fixes

eatkins:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/196/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I was able to jury rig a successful appveyor sbt build against this commit: https://ci.appveyor.com/project/eatkins/sbt/builds/19348872. I had to revert the super shell PR in that sbt build because I couldn't get scripted tests to work with a snapshot io after that PR. I suspect that the build will work with an IO 1.3.0-M3 built off of these commits.

The appveyor builds tend to hang if the swoval pathfinder implementation is used. I don't really understand why that is, but I'm going to continue investigating while we move forward. The scripted tests do work on my windows vm so I don't think there is a huge issue. I even ran the tests with a pathfinder that ran both the nio and swoval implementations and if there was a diff, it printed it, but I didn't see any significant diffs (apart from some diffs that would be explained by concurrency).

eed3si9n (approved):

Issue #164, opened 2018-06-05
IO.zip nullpointerexception if outputfile does not have path info in it


aappddeevv:
Using `IO.zip(...files.., file("blah.zip"))` fails if blah.zip does not have path information in it.
```sh
()
[error] java.lang.NullPointerException
[error] 	at sbt.io.IO$.createDirectory(IO.scala:221)
[error] 	at sbt.io.IO$.archive(IO.scala:509)
[error] 	at sbt.io.IO$.zip(IO.scala:498)
```
Changing blah.zip to `./blah.zip` works.

eed3si9n:
@aappddeevv Thanks for the report. Would you be interested in fixing this bug?

mrveera:
By adding null check on below line can fix this issue.
https://github.com/sbt/io/blob/896c6f5f38e326f43ecf8d3f2eb8d68ec7a880e3/io/src/main/scala/sbt/io/IO.scala#L607

PR #197, opened 2018-10-08
Fixes sbt/io#164 null check added before creating directory

mrveera:
Fixes sbt/io#164

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -604,9 +604,9 @@ object IO {
     if (outputFile.isDirectory)
       sys.error("Specified output file " + outputFile + " is a directory.")
     else {
-      val outputDir =  Option(outputFile.getParentFile) match {
-        case Some(parentFile) => parentFile
-        case None             => new File(".")
+      val outputDir = outputFile.getParentFile match {
+        case null       => new File(".")
+        case parentFile => parentFile
       }
       createDirectory(outputDir)
       withZipOutput(outputFile, manifest) { output =>
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -604,7 +604,10 @@ object IO {
     if (outputFile.isDirectory)
       sys.error("Specified output file " + outputFile + " is a directory.")
     else {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/197/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -604,7 +604,10 @@ object IO {
     if (outputFile.isDirectory)
       sys.error("Specified output file " + outputFile + " is a directory.")
     else {
-      val outputDir = outputFile.getParentFile
+      val outputDir =  Option(outputFile.getParentFile) match {
+        case Some(parentFile) => parentFile
+        case None             => new File(".")
+      }
```
eatkins, line 610: In this case, I'd probably avoid wrapping in Option and just do
```scala
val outputDir = outputFile.getParentFile match {
  case null        => new File(".")
  case parentFile  => parentFile
}
```
since there is no way for the null to propagate out of the match.

eatkins (changes requested):

io/src/main/scala/sbt/io/IO.scala
mrveera: Sure will do that.

eatkins (approved):

eatkins:
Hmm...the scalafmt tests are failing on 2.13.0-M4. Try running '++2.13.0-M4 scalafmt' and check the diff. This is confusing because I wouldn't expect scalafmt to depend on the scala version.

mrveera:
> Hmm...the scalafmt tests are failing on 2.13.0-M4. Try running '++2.13.0-M4 scalafmt' and check the diff. This is confusing because I wouldn't expect scalafmt to depend on the scala version.

I think they are formatting issues, Give it a minute they will pass soon

eatkins:
Merged. Thanks for the contribution @veera83372 !

PR #201, opened 2018-11-04
Fix anti-entropy bug

eatkins:
I noticed that sometimes my builds against the latest sbt nightly would
not re-trigger if my source file was touched while the test was running.
It turned out that the reason was that the condition that checked if the
file event occurred within the anti-entropy period was incorrect. I'm
not really sure why I wrote the initial condition this way.

The new test fails without the change to FileEventMonitor.scala.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/201/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #207, opened 2018-12-15
Remove broken JNA code for FreeBSD

alexdupre:
FreeBSD 12 changed the `struct stat` ABI, with the consequence that the sbt io code badly breaks and make the JVM to crash in very unexpected ways.

I have patches for the openjdk8 port, that will be committed soon, that adds millisecond resolution to the get/setLastModifiedTime functions, so this custom native code is not needed anymore.

This patch can be merged also in the 1.2.x branch.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Milli.scala
+++ b/io/src/main/scala/sbt/internal/io/Milli.scala
@@ -224,20 +224,6 @@ private object Linux32Milli extends PosixMilliIntUtim[Linux32] {
   }
 }
 
-private class FreeBSD64FileStat extends StatLong(120, 40, 48)
-private trait FreeBSD64 extends Library with Utimensat[Long] {
-  def stat(filePath: String, buf: FreeBSD64FileStat): Int
-}
-private object FreeBSD64Milli extends PosixMilliLongUtim[FreeBSD64] {
-  protected final val AT_FDCWD: Int = -100
-  protected final val UTIME_OMIT: Long = -2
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/207/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/internal/io/Milli.scala
```diff
@@ -224,20 +224,6 @@ private object Linux32Milli extends PosixMilliIntUtim[Linux32] {
   }
 }
 
-private class FreeBSD64FileStat extends StatLong(120, 40, 48)
```
eatkins, line 227: Mima seems to have problems with private classes and traits inside of an object. If you add the missing classes to the mimaBinaryIssueFilters setting, CI should pass.

alexdupre: Thanks, fixed.

cunei (approved):
Looks good, thanks @alexdupre. That reminds me that we can modify the Milli code to use regular Java calls on JDK10 or greater, now. Opened a related issue at https://github.com/sbt/io/issues/211

PR #208, opened 2018-12-15
avoid exposing private type in API

bbjubjub2494:
a bunch of methods in `Using` expose `OpenFile[T]` as a type. It took me a while to figure out that it was a subtype of `Using[File, T]` since `OpenFile` isn't exposed. I reckon changing the exposed type could improve UX. This is so since the first commit, so I don't know if there's a particular reason.

### DIFF
--- a/io/src/main/scala/sbt/io/Using.scala
+++ b/io/src/main/scala/sbt/io/Using.scala
@@ -69,9 +69,9 @@ object Using {
       def close(s: T) = closeF(s)
     }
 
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/208/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #212, opened 2018-12-16
Add dsl for specifying sources

eatkins:
It is somewhat clunky in sbt to declare a new source. One must write
something like:
watchSources += Source(sourceDirectory.value, "*.proto", NothingFilter, true)

This gets tedious and is somewhat hard to read. In this commit, I add a
simple dsl that allows us to build up soures:
watchSources += sourceDirectory.value / ** / *.proto
I think this is more readable and makes the intent more clear.

### DIFF
--- a/io/src/main/scala/sbt/io/Path.scala
+++ b/io/src/main/scala/sbt/io/Path.scala
@@ -18,12 +18,13 @@ import java.nio.file.{
 
 import com.swoval.files.FileTreeViews
 import com.swoval.functional.Filter
+import sbt.internal.io.Source
 import sbt.io.FileTreeView.AllPass
 
 import scala.collection.JavaConverters._
 import scala.collection.mutable
 
-final class RichFile(val asFile: File) extends AnyVal with RichNioPath {
+final class RichFile(val asFile: File) extends AnyVal with RichNioPath with DSLExtensions {
   def /(component: String): File = if (component == ".") asFile else new File(asFile, component)
 
   /** True if and only if the wrapped file exists.*/
@@ -86,6 +87,39 @@ final class RichFile(val asFile: File) extends AnyVal with RichNioPath {
 
   def withLinkOptions(linkOption: LinkOption*): LinkOptionPath =
     new LinkOptionPath(asPath, linkOption.toVector)
+}
+
+private[sbt] trait DSLExtensions extends Any {
+  def asFile: File
+  def /(g: PathFinder.**.type): Source =
+    new Source(asFile, AllPassFilter, NothingFilter, recursive = true)
+  def /(g: PathFinder.*.type): Source =
+    new Source(asFile, AllPassFilter, NothingFilter, recursive = false)
+  def /(filter: FileFilter): Source =
+    new Source(asFile, filter, NothingFilter, recursive = false)
+
+  def toSource: Source = {
+    val path = asFile.toPath
+    val (base, recursive, filter) = path.iterator.asScala
+      .foldLeft(
+        (path.getFileSystem.getRootDirectories.asScala.head, false, AllPassFilter: FileFilter)) {
+        case ((basePath, rec, f), part) if part.toString == "**" && !rec =>
+          (basePath, true, f)
+        case ((basePath, rec, f), part) if part.toString.startsWith("*.") && f == AllPassFilter =>
+          (basePath, rec, new ExtensionFilter(part.toString.drop(2)))
+        case ((basePath, rec, f), part) => (basePath.resolve(part), rec, f)
+      }
+    new Source(base.toFile, filter, NothingFilter, recursive)
+  }
+}
+
+final class RichSource(val s: Source) extends AnyVal {
+  def /(filter: FileFilter): Source =
+    new Source(s.base, filter, NothingFilter, recursive = s.recursive)
+  def &&(filter: FileFilter): Source =
+    new Source(s.base, s.includeFilter && filter, s.excludeFilter, s.recursive)
+  def --(filter: FileFilter): Source =
+    new Source(s.base, s.includeFilter -- filter, s.excludeFilter, s.recursive)
 }
 
 final class LinkOptionPath(p: NioPath, lo: Vector[LinkOption]) extends RichNioPath {
@@ -333,6 +367,11 @@ object PathFinder {
 
   def strict(files: Traversable[File]): PathFinder = apply(files)
 
+  case object **
+  import scala.language.dynamics
+  object * extends Dynamic {
+    def selectDynamic(name: String): ExtensionFilter = new ExtensionFilter(name)
+  }
 }
 
 /**
@@ -448,7 +487,7 @@ sealed abstract class PathFinder {
    * Create a PathFinder from this one where each path has a unique name.
    * A single path is arbitrarily selected from the set of paths with the same name.
    */
-  def distinct(): PathFinder = PathFinder { get().map(p => (p.asFile.getName, p)).toMap.values }
+  def distinct(): PathFinder = PathFinder { get().map(p => (p.getName, p)).toMap.values }
 
   /**
    * Constructs a string by evaluating this finder, converting the resulting Paths to absolute path strings,
--- a/io/src/main/scala/sbt/io/syntax.scala
+++ b/io/src/main/scala/sbt/io/syntax.scala
@@ -1,4 +1,5 @@
 package sbt.io
+import sbt.internal.io.Source
 
 private[sbt] trait Alternative[A, B] {
   def |(g: A => Option[B]): A => Option[B]
@@ -18,6 +19,8 @@ object syntax extends IOSyntax0 {
   type File = java.io.File
   type URI = java.net.URI
   type URL = java.net.URL
+  val * = PathFinder.*
+  val ** = PathFinder.**
 
   def uri(s: String): URI = new URI(s)
   def file(s: String): File = new File(s)
@@ -25,4 +28,5 @@ object syntax extends IOSyntax0 {
 
   implicit def fileToRichFile(file: File): RichFile = new RichFile(file)
   implicit def filesToFinder(cc: Traversable[File]): PathFinder = PathFinder.strict(cc)
+  implicit def sourceToRichSource(source: Source): RichSource = new RichSource(source)
 }
--- a/io/src/test/scala/sbt/io/GlobFilterSpec.scala
+++ b/io/src/test/scala/sbt/io/GlobFilterSpec.scala
@@ -1,7 +1,11 @@
 package sbt.io
 import java.io.File
+import java.nio.file.Files
 
 import org.scalatest.{ FlatSpec, Matchers }
+import sbt.internal.io.Source
+
+import scala.util.Properties
 
 class GlobFilterSpec extends FlatSpec with Matchers {
   "GlobFilter" should "work with *" in {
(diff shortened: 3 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/212/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
For context, I use this in a WIP topic branch:
https://github.com/eatkins/sbt/commit/8471e388095df81969aa3f85fd0af017d6dd98fd
https://github.com/eatkins/sbt/commit/afea1bb0b01aceaa0be02c8316c3bb81ebdeeda0

io/src/main/scala/sbt/io/Path.scala
```diff
@@ -88,6 +89,39 @@ final class RichFile(val asFile: File) extends AnyVal with RichNioPath {
     new LinkOptionPath(asPath, linkOption.toVector)
 }
 
+private[sbt] trait DSLExtensions extends Any {
+  def asFile: File
+  def /(g: PathFinder.**.type): Source =
```
eed3si9n, line 94: I think overloading `/` infix operator is too subtle.
If I am understanding this correctly, `dir / "somthing"` would return `File`, but `dir / **` would return `Source`?

eed3si9n (changes requested):

io/src/main/scala/sbt/io/Path.scala
eatkins: Yes, that's correct. I use the dsl heavily in this WIP commit:
https://github.com/eatkins/sbt/commit/b16600c6b1908f09daab98a3cbfc3c41496bf9aa

I'm not sure I agree that it's too subtle. For example, I think it's even more subtle that
```
watchSources ++= (sourceDirectory.value ** "*.scala").get
```
works due to the implicit conversion between files and sources.

That being said, is there an alternative syntax that you'd recommend? My main gripe is that I think that sources should be a first class citizen within sbt and right now they are a bit confusing to work with. My argument is essentially that
```
watchSources += baseDirectory.value / ** / *.scala
``` 
is significantly better than
```
watchSources += Source(baseDirectory.value, "*.scala", NothingFilter, recursive = true)
```
One alternative that I can think of that I don't love but is maybe less obtuse is
```
watchSources += baseDirectory.value.recursive.include(*.scala)
```

eatkins: I should add that I like the new syntax because it matches how I would describe the sources using globbing in bash or zsh. I think this makes the conceptual overhead of reading the build definitions lower even if the implementation is slightly harder to understand. FWIW, go to definition works (at least in intellij) so it's pretty easy to figure out what's going on.

io/src/main/scala/sbt/io/Path.scala
```diff
@@ -333,6 +367,11 @@ object PathFinder {
 
   def strict(files: Traversable[File]): PathFinder = apply(files)
 
+  case object **
+  import scala.language.dynamics
+  object * extends Dynamic {
```
eed3si9n, line 372: One of the first things I did when I started hacking on sbt/io was giving English name for `*` operator  e61821319b4e7c0074567d93e1428ad5a7a52aad.

Making `*` a value available to `build.sbt` with dynamic turned on would be confusing to Scala developers, since we'd be like "wait.. how does this work?"

io/src/main/scala/sbt/io/Path.scala
eed3si9n: > That being said, is there an alternative syntax that you'd recommend?

I'd suggest something like this:

```scala
watchSources += baseDirectory.value watchRecursive withExtension("*.scala")
```

io/src/main/scala/sbt/io/Path.scala
eatkins: So we went in the opposite direction :) This is opt-in since I wouldn't be deprecating the old syntax.

I don't really know where we should draw the line on confusing language features (and I agree this is one if you aren't familiar with it). I found the implicit conversion from string to extension filter, e.g.
```
includeFilter in unmanagedSources := "*.java" || "*.scala"
```
difficult to understand because the implicit definitions were hard to find back in the days when I used to develop exclusively with vim. At least with dynamic, you know there must be a class or object defined somewhere named "*" or "**" which is easy to find with grep. 

I suppose I could just publish the dsl is a plugin if you strongly object to this. My viewpoint is significantly influenced by using intellij all the time. Something like this might initially confuse me, but I'd figure it out quickly by doing a goto definition. If you aren't using intellij then this is a pain because you have to find the definition in the sbt source which could take a while.

eatkins:
I took out the TypedPath related commits and pushed a separate PR since the concerns are orthogonal.

io/src/main/scala/sbt/io/Path.scala
eatkins: Maybe we can compromise and provide both? I'd even be willing to put the `*` class that uses Dynamics behind an import, e.g. `import sbt.source.syntax._`.

eatkins:
Now that I think about it more, there is no reason this dsl needs to go in io and should be in sbt itself, if it all. I'll leave this open though until we agree on a syntax we're happy with or we decide not to provide any syntax.

io/src/main/scala/sbt/io/Path.scala
eed3si9n: With or without dynamic, I don't know `*` as a value is good idea in sbt since there's a bunch of uses of that.
Given that we already have a filter that converts from `"*.scala"`, the creation of extension filter should be more explicit like `withExtension("*.scala")`. Hopefully Googling "withExtension sbt" would turn up more result than "sbt "*""

eatkins: I guess, in my head, it's so obvious what `foo / ** / *.scala` should mean that I hadn't really considered searchability. Does searchability matter if the syntax is put behind an import since the author of the build file has explicitly opted in?

The reason that I feel strongly about this is that because I use sbt like a shell, I like the syntax to match the shell as closely as possible. It has much lower cognitive overhead, especially when I want to interactively change the build, e.g.
```
sbt:akka> set Compile / unmanagedResources / inputs += baseDirectory.value / ** / *.txt
[info] Defining Compile / unmanagedResources / inputs
[info] The new value will be used by Compile / unmanagedResources, watchSources
[info] Reapplying settings...
[info] Set current project to akka (in build file:/Users/ethanatkins/work/scratch/akka/)
```
The syntax that you're proposing is good for a static build file where it's only typed once, but is a bit onerous to type in an interactive session. This is also why I like avoiding quotes in the syntax (`baseDirectory.value / "*.scala"`).

eatkins: Perhaps the best compromise is to put the `*` syntax behind an import and add the explicit `withExtension` syntax at the sbt package level. To support the interactive use case, we'd make the `*` syntax available in the shell by default even though it wouldn't be available in build.sbt without import.

eed3si9n: > The reason that I feel strongly about this is that because I use sbt like a shell, I like the syntax to match the shell as closely as possible.

I am assuming that [Path Finders DSL](https://www.scala-sbt.org/1.x/docs/Paths.html#Path+Finders) was designed to mimic the feel of Bash or Zsh. The reason why I am calling this too subtle is because now you're proposing alternative DSL that kind of look the same, but does something else.

Having already spent the cognitive load on the sbt users to learn the Path Finders DSL, if we can extend the existing DSL, but somehow lift it into `Source` that would be the best of both worlds I think.

eatkins: Ok. I've thought a lot more about this and I think that the underlying issue is that `Source` and `PathFinder` describe completely different, though related things. The interface for `PathFinder` allows us to fetch paths, but tells us nothing about how it is doing the fetching or what it is fetching. Conversely, `Source` tells us what to fetch, but doesn't specify any implementation. This is why the `WatchSource` abstraction had to be added in the first place since `PathFinder` by definition cannot describe what sources we need to watch without eagerly fetching the files. It is possible to go `Source -> PathFinder`, but `PathFinder -> Source `is impossible in general.

Because the `Pathfinder` deliberately hides the implementation, it is impossible to apply certain kinds of optimizations on arbitrary PathFinders, e.g. proxying the requests through a cache. The `Source` api is much simpler and thus much more amenable to optimization. Moreover, watching files more or less requires the `Source` api, so that is permanently going to be part of the sbt api. It is simpler and less powerful that PathFinder which should make it easier for most users to reason about. Unfortunately without special syntax, the `Source` api is less natural to use than PathFinder when using the dsl.

The dsl for creating a `PathFinder` from a file is problematic because the inversion to `Source` isn't possible. The only way I can think of to fix this while preserving the PathFinder dsl is to partially break source compatibility and remove the implicit flag from the conversion from `File -> PathFinder` and add implicit conversions from `File -> Source` and `Source -> PathFinder`. To make this concrete, we'd have the following:
sbt 1.2.x
```scala
val pathFinder: PathFinder = baseDirectory.value
val files: Seq[File] = (pathFinder globRecursive AllPassFilter).get()
val alternativeFiles: Seq[File] = baseDirectory.value.allPaths.get()
```
sbt 1.3.x
```scala
val source: Source = baseDirectory.value
val files: Seq[File] = (source globRecursive AllPassFilter).get()
val alternativeFiles: Seq[File] = baseDirectory.value.allPaths.get()
```
My hope is that this would be possible in such a way that there are no binary incompatibilities and that the only source incompatibilities are places where people implicitly cast a `File` to a `PathFinder` directly. The idea is basically to just preserve the information that is lost when we convert File -> PathFinder so the dsl syntax does not actually need to change.

One possible alternative that I think would be less likely to cause source incompatibilities would be to deprecate the `Source` class and introduce a new class
```scala
final class PathFinderInput(val base: File, val filter: FileFilter, val recursive: Boolean) extends PathFinder
```
The reason I'd like to replace `Source` is that the name is singular but it actually describes a collection of `File`s. Moreover, I find the excludeFilter parameter unnecessary and confusing. While it is technically true that there are conditions that cannot be expressed without the excludeFilter, the vast majority of the time (and everywhere that I've seen it used in sbt), the exclude case can just be handled by replacing the inputFilter with inputFilter -- excludeFilter.

I intend to deprecate `watchSources` anyway for the new watch implementation and it will be trivial to convert between `PathFinderInput` and `Source`. Also `Source` is defined in the sbt.internal.io package and I'd imagine `PathFinderInput` would probably be in the sbt or sbt.io package. The `PathFinderInput` name is not my favorite, so I'm trying to think of a better one. I considered `Inputs`, but, in practice, I expect to mostly use collections of `PathFinderInput` and I find `val inputs = settingKey[Seq[Inputs]]("...")` a bit confusing due to the double pluralization implicit in `Seq[Inputs]`.

I will continue working on my sbt PR and will push some commits that implement this idea both with the `Source` class and with the new `PathFinderInput` class so we can see what the change would look like.

eed3si9n: `PathFinderInput` route sounds like a better plan, especially if the build user can opt into migrating from old `Source` style.

As per name is concerned, I naming it with the metaphor that's close to the build user's mental image would be good. How about something like `DirectoryWatch`, kind of giving hat tip to takari/directory-watcher? Or `WatchSet`.

> Also `Source` is defined in the `sbt.internal.io` package and I'd imagine `PathFinderInput` would probably be in the `sbt` or `sbt.io package`. 

Yea, it's an odd one. It's in `sbt.internal.io` but there's an alias for it called `sbt.WatchSource` - https://github.com/sbt/sbt/blob/v1.2.7/sbt/src/main/scala/sbt/Import.scala#L60-L61

eatkins: Yeah, I am proceeding down the PathFinderInput route and it's much cleaner. It shouldn't even break source compatibility for the vast majority of users (and it should be easy to fix where it does).

I don't want to put Watch in the name because PathFinderInput does not actually imply that it has anything to do with Watch. In addition to its usage in file watching, it is also needed to take advantage of the file system cache. In order to cache the file system tree, we need to be able to specify which portion of the file system tree we want to cache. Perhaps something like `FileTreeKey` or `FileSubTree` is more descriptive. Something like that would hopefully make it clear that it is an identifier for a subtree of the file system.

eatkins:
I replaced this with #216

PR #215, opened 2018-12-18
Don't automatically add path on error

eatkins:
There were two semantic bugs in swoval file tree views:

1) When the depth of the path search is set to -1, the file tree view is
   supposed to return a list that contains only the entry for the source
   root. For depth >= 0, the source root is not part of the results.

2) recursive traversal of the file system was halted if a subdirectory
   did not satisfy the filter.

I think that I added the exception handler in DefaultFileTreeView to
support getting the source root, but this caused problems elsewhere. I
fixed the upstream bug so that now DefaultFileTreeView should do the
right thing.

Similarly, the logic for first listing the results with AllPass and
subsequently applying the filter is unnecessary now that I fixed the
upstream bug.

Since there were not tests of these semantics, I added FileTreeViewSpec.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
+++ b/io/src/main/scala/sbt/internal/io/DefaultFileTreeView.scala
@@ -2,15 +2,12 @@ package sbt.internal.io
 
 import java.nio.file.{ NoSuchFileException, NotDirectoryException, Path }
 
-import com.swoval.files.{ TypedPath => STypedPath }
-import com.swoval.files.FileTreeViews
+import com.swoval.files.{ FileTreeViews, TypedPath => STypedPath }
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/215/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
This is a pre-req for the PathFinderInput idea in #212.

eed3si9n (approved):
Nice to see tests.

PR #217, opened 2019-01-16
Bump swoval

eatkins:
I found an off by one error that would erroneously evict a subdirectory
from monitoring if both the subdirectory and its parent were registered
with depth=0.  This is fairly rare in normal sbt builds since usually
only the base directory is registered with depth=0, but I discovered it
in a scripted test and it surely would have caused an issue for some
users.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/217/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #222, opened 2019-02-08
Bump swoval version

eatkins:
I discovered a nasty performance bug that impacted windows and linux in
swoval. Whenever the background watch process for the global file
repository detected that a directory was modified, it would rescan the
entire directory to determine what had changed. This could cause a huge
amount of io to happen if there were a lot of files in the modified
directory. It also was generally unnecessary because the directory
should have been updated via incremental file updates. This change does
make it slightly more likely that the cache will get out of sync with
the file system, but the cache is easily flushed and users can disable
it if they have any issues anyway.

### DIFF
--- a/io/src/main/scala/sbt/io/FileTreeRepository.scala
+++ b/io/src/main/scala/sbt/io/FileTreeRepository.scala
@@ -22,7 +22,7 @@ import scala.collection.mutable
  * Represents a path in the file system. It may cache some of the file attributes so that no disk
  * io is necessary to check, for example, whether the file is a directory or regular file.
  */
-trait TypedPath {
+private[sbt] trait TypedPath {
 
   /**
    * The underlying path that this represents.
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/222/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I found another performance bug so I have to make yet another swoval version. Things are looking really good with my manual testing on windows though now.

eed3si9n:
Nice. Good to catch these before 1.3.x.

eed3si9n (approved):

PR #223, opened 2019-02-08
Retries

eatkins:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/223/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -1383,4 +1372,36 @@ object IO {
     // see Java bug #6791812
     setModifiedTimeOrFalse(targetFile, math.max(last, 0L))
   }
+  private lazy val limit = {
+    val defaultLimit = if (scala.util.Properties.isWin) 100 else 1
```
eed3si9n: Given that current is 5 (?), 100 seems too high, and 1 seems too low.

eed3si9n (changes requested):

io/src/main/scala/sbt/io/IO.scala
eatkins: I think it's probably better to think in terms of the time it takes to retry rather than the number of retries. I (arbitrarily) chose 200ms as the maximum amount of time to spend retrying. The retries aren't really necessary on mac or linux because neither of those platforms generally lock files when they are accessing them (at least not in the context of how sbt uses the file system). I'll tune this parameter and see if CI reliably passes with a lower number.

eed3si9n: I think there's been issue with tempfs or networked stuff in the past, so I suggest we keep 5 as minimum.

eatkins: Ok. I have no problem with that.

eatkins: I switched it 10 for both platforms. I also set the sleep time to zero so that we yield control if there is another thread that is doing io on the file we're modifying. If there are no threads waiting, this shouldn't really matter.

eatkins: Windows sbt CI passed against a local version of io with these settings.

eed3si9n (approved):

PR #224, opened 2019-02-08
Convert glob base path to absolute path

eatkins:
If one does not convert the glob path to an absolute path, then the
filter that checks that the path starts with the base does not work. In
general, we want to use absolute paths as much as possible even though
there may be a memory overhead to doing so.

This was needed because the project/extra scripted test started failing
once I updated swoval to correctly handle relative paths. There was a
line like `(file("lib_managed") ** "*.jar").get` that was returning no
results even though there were jars in that directory. This was because
of the aforementioned filtering issue relative to the base directory.

### DIFF
--- a/io/src/main/scala/sbt/io/Glob.scala
+++ b/io/src/main/scala/sbt/io/Glob.scala
@@ -54,13 +54,17 @@ object Glob {
     override def hashCode: Int = f.hashCode
     override def toString: String = s"ConvertedFileFilter($f)"
   }
+  private implicit class PathOps(val p: NioPath) extends AnyVal {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/224/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

Issue #218, opened 2019-01-27
Drop 2.10 and (possibly) 2.11 support


eatkins:
It is becoming onerous to support 2.10 (it doesn't support SAM and there are compiler bugs in 2.10 that have been fixed in later versions, e.g. https://github.com/scala/bug/issues/8450). At this point, I don't really see the value proposition in continuing to publish 2.10 artifacts. I looked at the usages in maven and they are exclusively either part of sbt, zinc or typesafe (there are a few play related artifacts) or only publish a 2.12 target. Given that 0.13.x has been sunset, it seems reasonable to drop the 2.10 support, but perhaps I'm missing something.

eed3si9n:
Until https://github.com/sbt/zinc/pull/607 Zinc's test used to require the target Scala version. Another theoretical purpose of cross building IO was so the plugin author can migrate to new IO while they are still on sbt 0.13, and cross build, or use IO as plain library.
At this point I am open to dropping 2.10 and 2.11.

eed3si9n:
Fixed this in https://github.com/sbt/io/pull/225/commits/2fbfffacfc0e9161b7ea7ff74cb5d2f50f2e0c6f

PR #226, opened 2019-02-23
Refactor all of the glob/watch related interfaces

eatkins:
I came up with the FileTreeRepository interface roughly a year ago when
I was working on the close watch plugin. Since then, my understanding of
the problem has improved and I see that the interfaces that I defined
were not quite as clean as they could have been. The core problem that
the file cache tries to solve is that both listing _AND_ filtering of
files can be slow in sbt. Since the beginning, sbt has primarily used
java.io.File based apis for io. The problem with this is that if a user
wants to filter by file attributes, this will almost always lead to
extra io. That extra io will in turn lead to increased latency before
sbt can start compiling.  If we continue to use file based apis, this
will always be a problem.

I already started tackling this problem by introducing globs. The Glob
spec makes it possible to efficiently retrieve the source files.
Moreover in the current sbt version, I employ some dirty tricks to
return java.io.File instances that have their isRegularFile and
isDirectory methods overridden so that they use a memoized value
computed while we recursively fetch the sources. I realized, however,
that this was an ugly hack and still required the task author to think
about how downstream consumers of the source file might use them.

To improve the situation, I rethought the interface for fetching files.
Instead of effectively having a function from Glob => Seq[File], I want
to instead have a function from
(Glob, AttributeFilter) => Seq[(Path, Attributes)]
Path is better than file because it gives us the option of swapping out
the file system. Moreover, by returning the Attributes, we can
explicitly filter paths based on their attributes without having to do
any io.

This new api allowed me to simplify Glob so that instead of having a
filter from TypedPath => Boolean, it now just has a filter from
Path => Boolean. Most of the path methods do not do any io except for
Path.toRealPath(). This means that the Glob filter will almost always be
fast since it's likely just doing a file name check. Part of the
motivation for this change is that TypedPath was an awkward interface
since it combined file meta data (i.e. the path) with the file
attributes. To replace the functionality, I introduce two traits
SimpleFileAttributes and CustomFileAttributes. SimpeFileAttributes is
analogous to TypedPath without the getPath method while
CustomFileAttributes is a replacement for FileTreeDataView.Entry.

The reason for adding the CustomFileAttributes class and having it
implement the BasicFileAttributes interface is because it gives the
option of using it as a valid file attribute type in a custom file
system. The BasicFileAttributes interface partially mirrors the posix
stat struct which means to implement it, you pretty much need to stat
the file. We don't, have to eagerly implement it, however. The default
implementation implements SimpleFileAttributes (which are basically free
as part of the swoval directory listing api) and delegates the other
methods to a lazily computed default BasicFileAttributes instance.

With respect to listing files, I simplified the FileTreeView api.
Instead of returning a Seq[TypedPath], it returns a Seq[T] for some
generic type T. This allowed me to remove the largely redundant
FileTreeDataView interface. In practice, we mainly only care about
FileTreeView[(Path, SimpleFileAttributes)]
FileTreeView[(Path, CustomFileAttributes[T])]. The former maps closely
to the old default FileTreeView while the latter is similar to the old
FileTreeDataView.

I simplified the Observer interface and removed the redunant
CacheObserver interface. Instead of having the CacheObserver interface,
I make the various file tree repository instance implement
Observable[FileEvent[T]] where T is the cache value. Implementing an
Observer[FileEvent[T]] is very similar to implementing a CacheObserver
and we remove a redundant interface.

I formalized and improved the Registerable interface. When one
registers a glob with a FileTreeRepository, they probably want to
receive events specifically for that glob. I make that possible by
having register return Either[IOExceptin, Observable[T]] instead of
Either[IOException, Boolean]. This will be helpful in the watch refactor
where I dynamically register globs during task evaluation and want to
monitor them for watch events. Previously, I had to manage the filtering
to exclude, say, jar files that are were also registered with the
repository. With the new register interface, I can add a simple api in
sbt to mark which globs you want to watch for events in the task
definition. When we list a marked glob, we'll register it with the file
repository and add it to the other observables that we're already
monitoring.

Even though a lot of code had to change to implement the refactored
interfaces, most of the code changes were very obvious since there was a
pretty clear 1:1  mapping between the new and old interfaces. For the
most part, the tests would work with minimal tuning once they compiled.

Finally, I tried to ensure that all of the apis related to file
repository are marked private[sbt]. I'm still not sure if I've gotten
these interfaces right and I want the freedom to change them without
breaking binary compatibility. In subsequent commits (or a new PR), I
will move more of the implementation classes into sbt.internal.io.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
This failed

```
[info] - should detect many creations and deletions *** FAILED ***
[info]   sbt.io.FileTreeRepositorySpec.CountdownLatchOps(deletionLatch).await(sbt.io.FileTreeRepositorySpec.DEFAULT_TIMEOUT.*(10L)) was false (FileTreeRepositorySpec.scala:176)
```

eed3si9n:
Does this require update to https://github.com/sbt/sbt/pull/4539?

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1416/console">is in progress...</a>

eatkins:
Yeah. I need to update 
https://github.com/sbt/sbt/pull/4539 which I'm doing now. There is still a small degree of flakiness in the tests. I'm working on improving both swoval and the io tests to fix that.

eatkins:
The failure you saw last night could also have been caused by the jdk watch service bug. I'll switch the jdk8 test to use adoptjdk.

eatkins:
Also, does https://github.com/sbt/io/pull/225 mean we're dropping 2.10 and 2.11 support? I had to call .right on a number of Either instances because they didn't make Either right biased until 2.12.

typesafe-tools:
[//]: # (Refs: develop-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1416/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/develop">develop             </A> | sbt/sbt@8a641690f3068ab904a5792f605bd9e09ec31886
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@e35ac2f0913b22128b04b062b62af1937fc447c2
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@c685c0f385d5d98e45505b10af99dff0832f95ae
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@6d35f329b6b6be8da467eefc399ba9fa6f6725c0
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@62471c621503de366b550fee4ff2a0a712d967b4

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=develop&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

eed3si9n:
> Also, does #225 mean we're dropping 2.10 and 2.11 support? I had to call .right on a number of Either instances because they didn't make Either right biased until 2.12.

Yea. Let's drop 2.10 and 2.11 IO - https://github.com/sbt/io/issues/218

eatkins:
Cool. I keep finding flaky tests in CI, so I'm fixing them.

eatkins:
Ok. I fixed (or at least mitigated) the flakiness. The latest code has run about 10 times on appveyor and travis (between my personal and sbt ci instances) with no failures.

eatkins:
BTW, I am re-running the tests 100 consecutive times on my local machine every time I push. I keep finding flaky tests that way. That is why there are so many extra commits.

eatkins:
I finally got 100 consecutive runs to pass on my machine.

io/src/main/scala/sbt/internal/io/LazyFileAttributes.scala
```diff
@@ -0,0 +1,49 @@
...
+ * (http://www.apache.org/licenses/LICENSE-2.0).
+ */
+
+package sbt.internal.io
+import java.nio.file.attribute.{ BasicFileAttributes, FileTime }
+import java.nio.file.{ Files, Path }
+
+import scala.util.Try
+
+private[sbt] class LazyFileAttributes(path: Path,
```
eed3si9n: Could you add Scaladoc documenting the purpose and state of this class?
I am not sure if lazy is a good name for it, if it caches the result forever. Maybe it should be called `FileAttributesOnce`?

eatkins: Good point. I think I might back out of extending BasicFileAttributes for now and only implement SimpleFileAttributes.

io/src/main/scala/sbt/internal/io/Evaluate.scala
```diff
@@ -0,0 +1,8 @@
+package sbt.internal.io
+import scala.util.control.NonFatal
+
+private[io] object Evaluate {
```
eed3si9n: Is this class still needed? I can compile without it.

io/src/main/scala/sbt/internal/io/LazyFileAttributes.scala
eatkins: I was still thinking about implementing the nio FileSystem when I wrote this but I think that is overkill for 1.3.0.

io/src/main/scala/sbt/internal/io/TimeSource.scala
```diff
@@ -0,0 +1,47 @@
+package sbt.internal.io
+import scala.concurrent.duration._
+
+private[io] trait TimeSource {
```
eed3si9n: Could you provide documentation on why this is needed, especially as implicit?

eatkins: Sure. I got lazy because it was private[io]. The commit message in https://github.com/sbt/io/pull/226/commits/f7ecd379d470ac8f2251ce039cdb20d5243889dc sort of explains it, but I'll add scaladoc. The short of it is that using Deadline.now leads to flaky tests.

eatkins: https://github.com/sbt/io/compare/b86688d63868d82d264c10f6160ac8e29439d375..7e2770102359c6688868627f124c3306649c1f63

io/src/main/scala/sbt/internal/io/Observers.scala
```diff
@@ -0,0 +1,161 @@
...
+
+/**
+ * An object that monitors a file system. The interface is very similar to that provided by other
+ * libraries/frameworks, such as [[http://reactivex.io/intro.html rxJava]]. When it detects changes
+ * in the file system, it will invoke a set of user specified callbacks. The Observable also
+ * allows the user to add and removes paths to monitor.
+ *
+ * @tparam T the generic type of observable values
+ */
+private[sbt] trait Observable[+T] extends AutoCloseable {
```
eed3si9n: Is RxJava style APIs getting a promotion here by moving it to top level?
It might be helpful to note your current thinking about its shelf life.

eed3si9n (approved):

eed3si9n:
So this PR is scheduled for 1.3.0-RC1, but not M1?

eatkins:
Correct. I have https://github.com/sbt/sbt/pull/4539 working against the current io origin/develop branch. The other pr, https://github.com/sbt/sbt/pull/4512, needs some minor surgery so I'd rather incorporate these breaking changes into that one.

io/src/main/scala/sbt/internal/io/Observers.scala
eatkins: It still in the internal package and private sbt so I don’t really think of this as a promotion. I’m not sure if it will survive in the long run depending on how the replacement file system pans out.

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1495/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: develop-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1495/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/develop">develop             </A> | sbt/sbt@c2602e0d4441da02366fa68ad11592e8670ccccd
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@41d5d1cbde5a5791e517c8f3cc98a821ed790a0b
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@428505df46d093bbaec0265042eb7a63daae51c3
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@f6c267a208809df6a15d8ce8790e6ef0a99aeb37
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@48574c204117a91a47d10e19c52826ec4e5195c3

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=develop&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

io/src/main/scala/sbt/nio/Glob.scala
```diff
@@ -209,7 +209,8 @@ object Glob extends LowPriorityGlobOps {
     val sorted = globs.toSeq.sorted
     val needListDirectory: Path => Boolean = (path: Path) => {
       val filters = sorted.map(g => Glob(g.base, (0, g.range._2), AllPass).filter)
-      filters.exists(_.accept(path))
+      val a = path.resolve("a")
```
eed3si9n: What does `"a"` mean?

io/src/main/scala/sbt/nio/Glob.scala
```diff
@@ -123,10 +123,10 @@ object Glob extends LowPriorityGlobOps {
     }
     def \(component: String): Glob = this / component
     def glob(filter: FileFilter): Glob =
-      new GlobImpl(converter(repr), (0, 1), ConvertedFileFilter(filter))
+      new GlobImpl(converter(repr), (1, 1), ConvertedFileFilter(filter))
```
eed3si9n: Is this a semantic change from the PathFinder behavior?

io/src/main/scala/sbt/nio/Glob.scala
```diff
@@ -60,17 +61,284 @@ private[sbt] sealed trait ToGlob extends Any {
   def toGlob: Glob
 }
 object Glob extends LowPriorityGlobOps {
-  private[sbt] class ConvertedFileFilter(val f: FileFilter) extends PathNameFilter {
+  import scala.language.dynamics
```
eed3si9n: I'll link to the discussion we had in https://github.com/sbt/io/pull/212.

eed3si9n: Commit message says:

> Globs conceptually map more or less identically to the globs present in zsh or bash.

This is a bit of a stretch because glob wildcard should support `*`, including `*abc*`, `?`, `[co]`, and `[a-z]`. Linux also allows `[!x]` to mean negation. See [POSIX 2.13. Pattern Matching Notation]( http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_13).

io/src/main/scala/sbt/nio/Glob.scala
```diff
@@ -60,17 +61,284 @@ private[sbt] sealed trait ToGlob extends Any {
...
+
+  /**
+   * Used to construct recursive globs using the `/` method of [[GlobOps]]
+   */
+  case object **
```
eed3si9n: Generally not a huge fan of symbolic ops (as I've mentioned about https://github.com/sbt/io/commit/e61821319b4e7c0074567d93e1428ad5a7a52aad adding English name for `def **` in 2015), but having two different meanings for `**` would be bad design.

Note that `*` (`glob`) and `**` (`globRecursive`) methods already return Glob in sbt 1.3:

```scala
scala> val g = base * "*.scala"
g: sbt.io.Glob =
Glob(
  base = /Users/eed3si9n/work/akka,
  filter = ConvertedFileFilter(ExtensionFilter(scala)),
  depth = 0
)

scala> val g = base ** "*.scala"
g: sbt.io.Glob =
Glob(
  base = /Users/eed3si9n/work/akka,
  filter = ConvertedFileFilter(ExtensionFilter(scala)),
  depth = 2147483647
)

scala> val g = base globRecursive "*.scala"
g: sbt.io.Glob =
Glob(
  base = /Users/eed3si9n/work/akka,
  filter = ConvertedFileFilter(ExtensionFilter(scala)),
  depth = 2147483647
)
```

Whatever the DSL we add here I think needs to be consistent with the rewrite we have to do in https://www.scala-sbt.org/1.x/docs/Paths.html. The case where a new DSL might be needed is expressing the new range.

I would propose adding an object named `anyPath` for this purpose, which can be used as:

```scala
scala> val g = base glob (anyPath / anyPath / "*.scala")
```

eed3si9n (changes requested):

io/src/main/scala/sbt/nio/Glob.scala
eatkins: Good point. A more accurate sentence would have been "Globs are intended to map to posix pattern matching notation with some extended features, such as recursive globbing with '**' that are found in shells like zsh".

eatkins: I'll just take out the `**` and `*` variables. I can make a plugin if I really care. The main reason I liked them was that I could do things in the console `set watchTriggers += baseDirectory.value / ** / *.txt` and wouldn't have to type quotation marks. A nicer change might be to enhance the `set` command to automatically insert quotation marks if they are missing in certain situations. I find that it's also annoying to have to type things like `set version := "1.3.0"` rather than `set version := 1.3.0`.

eatkins: I still dislike `baseDirectory.value ** AllPassFilter` and think that `baseDirectory.value / **` is better. Perhaps the solution really should be to just remove the symbolic operators entirely in sbt 2. Also, even though `baseDirectory.value / ** / *.txt` is kind of nice, it is confusing that you can't do `baseDirctory.value / ** / foo*.txt`. I guess adding special syntax isn't worth it when it's so limited.

eatkins: I just discovered this: https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String). I think this will do what I want.

io/src/main/scala/sbt/nio/Glob.scala
eatkins: I think the way to go is just to have `def /(string: String)` in GlobOps and use the https://docs.oracle.com/javase/7/docs/api/java/nio/file/FileSystem.html#getPathMatcher(java.lang.String) so that `base / "**"`, `base / "*" / "*.txt"`,` base / "*/*.txt"` are all valid.

eatkins: Side note: the only reason that I proposed the `**` was because the `def /(**.type)` method only applied to `Path` and `Glob`. It could not be applied to `File`, so while `baseDirectory.value ** AllPassFilter` worked, `baseDirectory.value / **` did not. The `Glob` instances returned by `baseDirectory.value ** AllPassFilter` and `baseDirectory.value.toPath / **` were the same on this branch so they were not actually significantly conceptually different. The point is moot though because I've already removed the `**` and `*` objects on my local brach. After I have lunch, I'm going to switch the glob parser to use the built in nio glob parser.

io/src/main/scala/sbt/nio/Glob.scala
eatkins: Yes and no. It is semantically different from 1.3.0-M2 when using `Glob` apis but is semantically identically when using PathFinder because we have:
```scala
  private[sbt] final class GlobPathFinder(val glob: Glob) extends PathFinder {
    override def get(): Seq[File] = {
      if (glob.range._1 == 0 && glob.range._2 == 0) {
        glob.base.toFile :: Nil
      } else if (glob.range._2 > 1) {
        val files = new java.util.LinkedHashSet[File].asScala
        **if (glob.pathFilter.accept(glob.base)) files.add(glob.base.toFile)**
        Path.defaultDescendantHandler(glob.base.toFile, glob.toFileFilter, files, glob.range._2)
        files.toIndexedSeq
      } else {
        Path.defaultChildHandler(glob.base.toFile, glob.toFileFilter)
      }
    }
  }
```
The line marked with ** was added to make `globRecursive` consistent with the old PathFinder api but the non recursive case never used the range of the glob anyway.

io/src/main/scala/sbt/nio/Glob.scala
eatkins: I should add a comment. Say I have a Glob`"/foo/bar/*/*/*.txt"` (which would have the representation `Glob(Paths.get("/foo.bar", (3, 3), "*.txt")`) and I want to check if the path `/foo/bar/baz/` needs to be listed. To do this, I simplify the glob to `Glob(paths.get("/foo/bar/", (0, 3), AllPass). I then create the path `/foo/bar/baz/a`. Because that path has depth 2, it is included in the range which means that it is possible that there are children of `/foo/bar/baz` that might be accepted by "/foo/bar/*/*/*.txt". If, on the other hand, the path was `/foo/bar/baz/buzz/banana`, the depth `/foo/bar/baz/buzz/banana/a` is 4 and it would be rejected by the simplified glob so we won't list that path.

eatkins: Writing that comment, I realized that I could rewrite this:
`sorted.exists(g => path.startsWith(g.base) && base.relativize(path).getNameCount < g.range._2).

io/src/main/scala/sbt/nio/Glob.scala
eed3si9n: Given `base` is `java.io.File`, would it require `toGlob` like:

```scala
base.toGlob / "*" / "*" / "*.scala"
```

?

eatkins: Yeah. Or base.toPath.

eatkins: Yeah. Or base.toPath.

io/src/main/scala/sbt/nio/Glob.scala
```diff
@@ -0,0 +1,518 @@
...
+  /**
+   * The filter to apply to elements found in the file system subtree.
+   * @return the filter.
+   */
+  def pathFilter: PathFilter
+
+}
+private[sbt] sealed trait GlobBuilder[G] extends Any {
+  def /(component: String): G
+  def \(component: String): G
```
eed3si9n: What's `def \` do?

eatkins: The idea is that on windows the path separator is `\` rather than `/` like on osx and windows. This trait is confusing and I was only using it to convert files to globs (at one point I was also using it to convert Paths to Globs) so I went ahead and just made a single concrete class FileGlobBuilder with `def \(component: String): Glob` and `def \(component: String): Glob`. The reason for FileGlobBuilder is to enable the PathFinder dsl to be used to specify globs. I'm increasingly unsure if that is a good idea or not. Perhaps we shouldn't mix and match the apis.

eed3si9n: Yea. Path on sbt should be expressed using `/` only.

eatkins: Ok, for legacy reasons, it needs to persist in the file based apis for binary compatibility, but I'll remove the `\` methods from GlobOps.

io/src/main/scala/sbt/nio/Glob.scala
eatkins: I believe that the sbt globs should be very close to shell globs with this commit: https://github.com/sbt/io/pull/226/commits/880ed6efa85df2512acb2ed618e22a44447d4041

io/src/test/scala/sbt/nio/GlobSyntaxSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+import sbt.io.syntax._
+import sbt.io.AllPassFilter
+import sbt.nio.syntax._
+
+import scala.util.{ Properties, Try }
+
+class GlobSyntaxSpec extends FlatSpec {
+  import GlobSyntaxSpec._
+  "path builders" should "use `*` and `**` objects" in {
+    val path = Paths.get("/foo/bar").toAbsolutePath
```
eed3si9n: It makes sense for a `Glob`'s `/` to return a new `Glob`, but does it make sense for `java.nio.file.Path`?
I'd expect `base / "foo"` to return another `Path`.

io/src/main/scala/sbt/nio/Glob.scala
eatkins: Should be resolved by https://github.com/sbt/io/pull/226/commits/f7e9d5fba02596aac99dfa2f75a3a5052449fd6a

io/src/test/scala/sbt/nio/GlobSyntaxSpec.scala
eatkins: And now we're back to the reason why I wanted `**` and `*`.

eatkins: Writing a dsl that satisfies everyone is hard.

eatkins: What if base / "foo" returned `Path with Glob`? I kind of like that.

eatkins: The more I think about this, the less certain I am that Glob is defined correctly. Conceptually, a file tree glob is a sequence of filters that are applied at successive levels of the file tree. What is tricky is that Path is effectively a special case of Glob. You do make a good point that '/' shouldn't lift a Path to a Glob.  I think the proper solution is a custom file system. Since there isn't really time for that, I think the best way forward is to private[sbt] the methods in the Glob base trait and just provide the `/` operator in GlobOps. Globs could then be instantiated using the PathFinder dsl and using the stringToGlob implicit conversion function that I already added. While I don't love this solution, it buys the most time to properly solve the problem while still having at least a somewhat ergonomic api, e.g. `set watchTriggers += s"${baseDirectory.value}/**/*.scala"` is acceptable. I prefer the string based api to both `set watchTriggers += baseDirectory.value ** "*.scala"` or `set watchTriggers += baseDirectory.value globRecursive "*.scala"` because it is pretty unambiguous what the intent is. The main drawback to the string based approach is that it isn't platform independent due to the file separator character differences between windows and posix systems.

eed3si9n: Maybe in another PR you could treat `"/"` as path separator regardless of the operating system.

eed3si9n: For reference, here's better-files https://github.com/pathikrit/better-files#globbing:

```scala
val dir = "src"/"test"
val matches: Iterator[File] = dir.glob("*.{java,scala}")
```

(except their `glob` means our `globRecursive` by default)

eatkins:
I've introduced a new syntax for `Glob` in https://github.com/sbt/io/pull/226/commits/b9292f1ff1aac6e0dc4d2d28962bdd5dcaf7e0dd
and
https://github.com/sbt/io/pull/226/commits/9981e1950bcf0ad557745ba04bf520d957aa37fd
I left the old syntax around to avoid the work of actually porting to the new syntax, which is relatively trivial, but still work that I don't want to throw out if the new syntax is discarded.

You can see in the tests added by https://github.com/sbt/io/pull/226/commits/b9292f1ff1aac6e0dc4d2d28962bdd5dcaf7e0dd what use cases are supported by the new `Glob`. In https://github.com/sbt/io/pull/226/commits/9981e1950bcf0ad557745ba04bf520d957aa37fd, I show how we can add implicit conversions to add a nice glob based dsl. Crucially, the method for lifting a `Path` to a `Glob` requires a relative `Glob` on the right hand side. I yet again added `**` and `*` but they are just aliases for objects called `Recursive` and `AnyPath`. I feel very strongly that the aliases are worth it while recognizing (and largely agreeing with) the general allergy to symbolic operators. I did not use dynamics to support, e.g. `*.txt`.

eed3si9n (approved):
LGTM.
Thanks for all your work on this!

eatkins:
Ok. Cool. I need to migrate the existing apis to the new syntax so I'll need to add a commit that makes the migration. I'll probably get that done on Sunday along with the long awaited sbt update.

eatkins:
@eed3si9n This is ready to go now from my perspective, but you may want to look over the new apis defined in https://github.com/sbt/io/blob/a1a9fc1e534b8181c9051a960d744ef65f3c0991/io/src/main/scala/sbt/nio/file/Glob.scala once more to make sure that we're on the same page.

eed3si9n:
@eatkins To clarify from user's perspective, could we pair this PR with a PR to https://github.com/sbt/website/blob/develop/src/reference/02-DetailTopics/02-Configuration/10-Paths.md? (We can make `1.x-beta` branch off of `develop`) We need release notes anyway for RC-1.

eatkins:
Sounds good. I'm going to get the sbt pr up to run dbuild first, but that will be the next step.

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1552/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: https://github.com/eatkins/sbt/pull/1-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1552/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/tree/https://github.com/eatkins/sbt/pull/1">https://github.com/eatkins/sbt/pull/1</A> | 
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@dc94df28346eae2a6d1486ba59be6aeb96d9c3ce
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@eae491f96e51670dfb8b2b3fd516bdb43d55692d
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@f073dde812848ef443619578f55e290f3863bcaa
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=https://github.com/eatkins/sbt/pull/1&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1553/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1553/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@a5ab32c775af39cd44a8b97fee2c9f7b70588d30
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@dc94df28346eae2a6d1486ba59be6aeb96d9c3ce
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@eae491f96e51670dfb8b2b3fd516bdb43d55692d
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@f073dde812848ef443619578f55e290f3863bcaa
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1567/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1567/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@a5ab32c775af39cd44a8b97fee2c9f7b70588d30
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@eae491f96e51670dfb8b2b3fd516bdb43d55692d
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@f073dde812848ef443619578f55e290f3863bcaa
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1568/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1568/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@a5ab32c775af39cd44a8b97fee2c9f7b70588d30
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@eae491f96e51670dfb8b2b3fd516bdb43d55692d
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@f073dde812848ef443619578f55e290f3863bcaa
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

eed3si9n:
```
[sbt212-librarymanagement] [info] OfflineModeSpec:
[sbt212-librarymanagement] [info] Offline update configuration
[sbt212-librarymanagement] [info] - should reuse the caches when offline is enabled *** FAILED ***
[sbt212-librarymanagement] [info]   java.nio.file.DirectoryNotEmptyException: ./target/ivyhome/cache
[sbt212-librarymanagement] [info]   at sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:242)
[sbt212-librarymanagement] [info]   at sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:108)
[sbt212-librarymanagement] [info]   at java.nio.file.Files.deleteIfExists(Files.java:1165)
[sbt212-librarymanagement] [info]   at sbt.io.IO$.$anonfun$delete$2(IO.scala:564)
[sbt212-librarymanagement] [info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.liftedTree2$1(Retry.scala:38)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.impl$1(Retry.scala:38)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.apply(Retry.scala:50)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.apply(Retry.scala:24)
[sbt212-librarymanagement] [info]   at sbt.io.IO$.delete(IO.scala:555)
```

The stacktrace shows sbt.delete and NIO, so maybe this is legit?

eed3si9n:
```
[sbt212-librarymanagement] [info] *** 22 TESTS FAILED ***
[sbt212-librarymanagement] [error] Failed: Total 113, Failed 22, Errors 0, Passed 91
[sbt212-librarymanagement] [error] Failed tests:
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.OfflineModeSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.IvyRepoSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.CachedResolutionSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.ModuleResolversTest
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.MergeDescriptorSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.FrozenModeSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.IvyResolutionSpec
[sbt212-librarymanagement] [error] 	sbt.internal.librarymanagement.InclExclSpec
[sbt212-librarymanagement] [error] (lmIvy / Test / test) sbt.TestsFailedException: Tests unsuccessful
```

It's entirely possible that it's basically some race condition that it's exposing.

eed3si9n:
https://github.com/sbt/librarymanagement/pull/300

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1569/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1569/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@a5ab32c775af39cd44a8b97fee2c9f7b70588d30
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@eae491f96e51670dfb8b2b3fd516bdb43d55692d
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@18f4fc75e5b307bb8afef49865be23cff57c9111
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -551,16 +552,16 @@ object IO {
...
   /** Deletes `file`, recursively if it is a directory. */
-  def delete(file: File): Unit = retry {
+  def delete(file: File): Unit = Retry {
     try {
-      FileTreeView.DEFAULT.list(Glob(file, AllPassFilter, 0)).foreach {
-        case dir if dir.isDirectory => delete(dir.toPath.toFile)
-        case f                      => f.toPath.toFile.delete()
+      FileTreeView.default.list(Glob(file, AnyPath)).foreach {
+        case (dir, attrs) if attrs.isDirectory => delete(dir.toFile)
+        case (f, _)                            => Files.deleteIfExists(f)
```
eed3si9n, line 559: https://jenkins.scala-sbt.org:8592/job/sbt-validator/1569/consoleFull

```
[sbt212-librarymanagement] [info] - should reuse the caches when offline and cached resolution are enabled *** FAILED ***
[sbt212-librarymanagement] [info]   java.nio.file.DirectoryNotEmptyException: ./target/OfflineModeSpec/ivyhome/cache
[sbt212-librarymanagement] [info]   at sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:242)
[sbt212-librarymanagement] [info]   at sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:108)
[sbt212-librarymanagement] [info]   at java.nio.file.Files.deleteIfExists(Files.java:1165)
[sbt212-librarymanagement] [info]   at sbt.io.IO$.$anonfun$delete$2(IO.scala:564)
[sbt212-librarymanagement] [info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.liftedTree2$1(Retry.scala:38)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.impl$1(Retry.scala:38)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.apply(Retry.scala:50)
[sbt212-librarymanagement] [info]   at sbt.internal.io.Retry$.apply(Retry.scala:24)
[sbt212-librarymanagement] [info]   at sbt.io.IO$.delete(IO.scala:555)
```

The error is `java.nio.file.DirectoryNotEmptyException: ./target/OfflineModeSpec/ivyhome/cache`. Does this point to the issue of caching?

eed3si9n (changes requested):

io/src/main/scala/sbt/io/IO.scala
eatkins: No, the issue is that FileTreeView.list(glob: Glob) didn't handle relative globs correctly. I both fixed the issue with relative globs and also changed this line from `FileTreeView.default.list(Glob(file, AnyPath)).foreach {` to `FileTreeView.default.list(file.toPath).foreach {` which should be semantically equivalent but has a simpler implementation. The failing librarymanagement test passes when using an io version incorporating the latest updates on my machine.

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1572/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1572/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@7e80be7a020320252085695f39f93e9bcf0fdb42
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@a365362e35d56e0b578b2f6b8455f068169a8a57
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@18f4fc75e5b307bb8afef49865be23cff57c9111
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1573/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1573/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@7e80be7a020320252085695f39f93e9bcf0fdb42
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@a365362e35d56e0b578b2f6b8455f068169a8a57
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@18f4fc75e5b307bb8afef49865be23cff57c9111
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1574/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1574/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@9bf8076e22f30589942d0ccb1baf9d0abd7003b0
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@d1c6f0b1487f90f107e9fd954f52daa15b6c373a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@18f4fc75e5b307bb8afef49865be23cff57c9111
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1575/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1575/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@58c377e943040d94f8b831c85ca1e91d4e26a37f
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@d1c6f0b1487f90f107e9fd954f52daa15b6c373a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

eatkins:
I think the validator failures are due to coursier, not my changes. The validator fails trying to run sbtProj/test on ServerSpec:
```
[sbt212:error] Downloading https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom.sha1
[sbt212:error] Downloading https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212:error] Downloaded https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212:error] Downloaded https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom.sha1
[sbt212:error] Downloading https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom.sha1
[sbt212:error] Downloading https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212:error] Downloaded https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom.sha1
[sbt212:error] Downloaded https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212:error] Downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
[sbt212:error] Downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml.sha1
[sbt212:error] Downloaded https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
[sbt212:error] Downloaded https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml.sha1
[sbt212] [[35mbuild-1[0m] waiting for the server...
[sbt212] [warn] 
[sbt212] [warn] 	Note: Unresolved dependencies path:
[sbt212] [error] sbt.librarymanagement.ResolveException: Error downloading org.scala-sbt:sbt:1.3.0-bin-20190430T161107
[sbt212] [error]   Not found
[sbt212] [error]   not found: /home/jenkinssbt/.ivy2/local/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
[sbt212] [error]   not found: https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212] [error]   not found: https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212] [error]   not found: https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
[sbt212] [error] 	at lmcoursier.CoursierDependencyResolution.unresolvedWarningOrThrow(CoursierDependencyResolution.scala:227)
[sbt212] [error] 	at lmcoursier.CoursierDependencyResolution.$anonfun$update$30(CoursierDependencyResolution.scala:196)
[sbt212] [error] 	at scala.util.Either$LeftProjection.map(Either.scala:573)
[sbt212] [error] 	at lmcoursier.CoursierDependencyResolution.update(CoursierDependencyResolution.scala:196)
[sbt212] [error] 	at sbt.librarymanagement.DependencyResolution.update(DependencyResolution.scala:60)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.resolve$1(LibraryManagement.scala:87)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$12(LibraryManagement.scala:137)
[sbt212] [error] 	at sbt.util.Tracked$.$anonfun$lastOutput$1(Tracked.scala:69)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$20(LibraryManagement.scala:150)
[sbt212] [error] 	at scala.util.control.Exception$Catch.apply(Exception.scala:228)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11(LibraryManagement.scala:150)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11$adapted(LibraryManagement.scala:131)
[sbt212] [error] 	at sbt.util.Tracked$.$anonfun$inputChanged$1(Tracked.scala:150)
[sbt212] [error] 	at sbt.internal.LibraryManagement$.cachedUpdate(LibraryManagement.scala:164)
[sbt212] [error] 	at sbt.Classpaths$.$anonfun$updateTask0$5(Defaults.scala:2842)
[sbt212] [error] 	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
[sbt212] [error] 	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
[sbt212] [error] 	at sbt.std.Transform$$anon$4.work(Transform.scala:67)
[sbt212] [error] 	at sbt.Execute.$anonfun$submit$2(Execute.scala:270)
[sbt212] [error] 	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:19)
[sbt212] [error] 	at sbt.Execute.work(Execute.scala:279)
[sbt212] [error] 	at sbt.Execute.$anonfun$submit$1(Execute.scala:270)
[sbt212] [error] 	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:178)
[sbt212] [error] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:37)
[sbt212] [error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[sbt212] [error] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[sbt212] [error] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[sbt212] [error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[sbt212] [error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[sbt212] [error] 	at java.lang.Thread.run(Thread.java:748)
[sbt212] [error] (update) sbt.librarymanagement.ResolveException: Error downloading org.scala-sbt:sbt:1.3.0-bin-20190430T161107
[sbt212] [error]   Not found
[sbt212] [error]   not found: /home/jenkinssbt/.ivy2/local/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
[sbt212] [error]   not found: https://repo1.maven.org/maven2/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212] [error]   not found: https://repo.scala-sbt.org/scalasbt/maven-snapshots/org/scala-sbt/sbt/1.3.0-bin-20190430T161107/sbt-1.3.0-bin-20190430T161107.pom
[sbt212] [error]   not found: https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-sbt/sbt/1.3.0-bin-20190430T161107/ivys/ivy.xml
```
When I run sbtProj/test the ServerSpec tries to resolve 1.3.0-SNAPSHOT rather than the bin version and everything works fine.

eed3si9n:
@eatkins Yes. server test now requires -SNAPSHOT to be published locally - https://github.com/sbt/sbt/commit/f999f6a62ea89fc8292e4be20564652b3757a04e

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1576/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1576/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@58c377e943040d94f8b831c85ca1e91d4e26a37f
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@d1c6f0b1487f90f107e9fd954f52daa15b6c373a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1582/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1582/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@58c377e943040d94f8b831c85ca1e91d4e26a37f
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@d1c6f0b1487f90f107e9fd954f52daa15b6c373a
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1583/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1583/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@6d4a6457c35ff086df3eec97cfe30454cdb0384e
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@b630b5531d4554dcf6556970cea9cbeb3ff8a449
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1586/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1586/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@2a4c7d1204c54c3d0050482196948136772e418e
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@b630b5531d4554dcf6556970cea9cbeb3ff8a449
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1587/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1587/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@2a4c7d1204c54c3d0050482196948136772e418e
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@b630b5531d4554dcf6556970cea9cbeb3ff8a449
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

typesafe-tools:
A validation involving this pull request <a href="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1588/console">is in progress...</a>

typesafe-tools:
[//]: # (Refs: pull/4627/head-develop-pull/226/head-develop-develop)
The validator <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/1588/">has checked</A> the following projects, tested using dbuild, projects built on top of each other.

  sbt               | <A HREF="https://github.com/sbt/sbt/pull/4627">pull/4627/head      </A> | sbt/sbt@2a4c7d1204c54c3d0050482196948136772e418e
  zinc              | <A HREF="https://github.com/sbt/zinc/tree/develop">develop             </A> | sbt/zinc@b6fc35f27615f129f91de9ac14400101fb8b9b7e
  io                | <A HREF="https://github.com/sbt/io/pull/226">pull/226/head       </A> | sbt/io@b630b5531d4554dcf6556970cea9cbeb3ff8a449
  librarymanagement | <A HREF="https://github.com/sbt/librarymanagement/tree/develop">develop             </A> | sbt/librarymanagement@984de6fe7281eb715ad270245f7c6eb190f23def
  util              | <A HREF="https://github.com/sbt/util/tree/develop">develop             </A> | sbt/util@3568b13a2e257f33957dd32c51c56ce9c378c9f9

<A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild?sbt=pull/4627/head&zinc=develop&io=pull/226/head&librarymanagement=develop&util=develop&deploy_to_bintray=false&run_sbt_community=false">(restart)</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -551,16 +552,16 @@ object IO {
...
-        case f                      => f.toPath.toFile.delete()
+      FileTreeView.default.list(file.toPath).foreach {
+        case (dir, attrs) if attrs.isDirectory => delete(dir.toFile)
+        case (f, _)                            => Files.deleteIfExists(f)
       }
     } catch {
       case _: NotDirectoryException =>
     }
-    file.delete()
+    Files.deleteIfExists(file.toPath)
```
eed3si9n, line 564: Maybe Windows is not ready for `Files.deleteIfExists(file.toPath)`?

eatkins: No. I reverted that and it made no difference. Based on the error message, underlying problem is that ScriptedTests is trying to delete a file that is still in use by the forked scripted process. The question is why? I also believe that somehow scripted is leaking classloaders because I've noticed a marked increase in the frequency off my running out of metaspace when running scripted in a continuous build.

eed3si9n: Looks like it was part of https://github.com/sbt/sbt/pull/3151, which allows scripted to reuse sbt instance.

eed3si9n: https://github.com/sbt/sbt/blob/bc21db28646f0b4a4484cdd22b6a4c4596cafde1/scripted-sbt-redux/src/main/scala/sbt/scriptedtest/ScriptedTests.scala#L329-L332

```scala
          // Run the test and delete files (except global that holds local scala jars)
          val result = runOrHandleDisabled(label, tempTestDir, runTest, buffer)
          IO.delete(tempTestDir.*("*" -- "global").get)
          result
```

The code has existed for a while, so what's newly holding on to the `project\target\config-classes\$17a69f0703746b7bed64$.class`?

PR #234, opened 2019-05-02
Don't return empty files in parseClasspath

eatkins:
Zinc will pass in an empty string into this method if a system property
isn't set. It doesn't make sense to handle the empty string as a
relative file. I noticed this because sbt now prints a warning if one
tries to list a relative glob (which is generated by the pathfinder dsl
in zinc).

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -591,6 +591,7 @@ object IO {
 
   /**
    * Creates a zip file.
+   *
    * @param sources The files to include in the zip file paired with the entry name in the zip.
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/234/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #235, opened 2019-05-02
Io bug fixes

eatkins:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/235/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -551,17 +551,23 @@ object IO {
...
+   * delete the file (or directory) if any errors occur.
+   */
   def delete(file: File): Unit = Retry {
     try {
       FileTreeView.default.list(file.toPath).foreach {
         case (dir, attrs) if attrs.isDirectory => delete(dir.toFile)
-        case (f, _)                            => Files.deleteIfExists(f)
+        case (f, _) =>
+          try Files.deleteIfExists(f)
+          catch { case _: IOException => }
```
eed3si9n, line 564: > The legacy behavior of IO.delete is that it leaves behind any files in
the tree that it is unable to delete. This is because File.delete only
throws if an AccessDeniedException occurs. On windows, it is not unusual
for deletions to fail because a jar or class file is in use somewhere.

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/File.html#delete()

ok. It's yuck, but it makes sense.

eed3si9n (approved):

eatkins:
The scripted tests succeeded on my vm after I published these changes locally.

PR #240, opened 2019-05-16
Fix glob conversion from certain file filters

eatkins:
Issue https://github.com/sbt/sbt/issues/4687 reports that sbt 1.3.0-RC1
does not work correctly with the sbt header plugin. This is because the
plugin uses the Defaults.collectFiles method that was changed to use
globs. There was an issue with the base path + file filter conversion
where if the filter was, say, AllPassFilter && !HiddenFileFilter, this
actually got translated to $baseDiretory/**/<null> (where <null>
indicates the NoPath filter). The primary cause of this issue was an
incorrect conversion from HiddenFileFilter to AnyPath. Hidden files
are automatically excluded by globs so the HiddenFileFilter is actually
equivalent to the NoPath filter. I also added the NotHiddenFileFilter
object to be the inverse of the HiddenFileFilter because
-HiddenFileFilter comes up in a lot of places so I think it makes sense
to handle it directly.

### DIFF
--- a/io/src/main/scala/sbt/internal/nio/Globs.scala
+++ b/io/src/main/scala/sbt/internal/nio/Globs.scala
@@ -66,9 +66,9 @@ private[sbt] object Globs {
             case _                                           => None
           }
         }
-      case AllPassFilter | HiddenFileFilter => Some(AnyPath)
-      case NothingFilter                    => Some(NoPath)
-      case _                                => None
+      case AllPassFilter | NotHiddenFileFilter => Some(AnyPath)
+      case NothingFilter | HiddenFileFilter    => Some(NoPath)
+      case _                                   => None
     }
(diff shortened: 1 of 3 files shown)


eatkins:
Fixes https://github.com/sbt/sbt/issues/4687

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/240/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
That was fast.

eed3si9n (approved):

PR #241, opened 2019-05-29
Fix extension filter

eatkins:
I noticed that sbt 1.3.0 had problems with the twirl template compiler
and I tracked it down to the template compiler calling effectively
(sourceDirectories `**` "*.template.scala").get.

This created an instance ofExtensionFilter with the extension "template.scala". The problem was
that ExtensionFilter had an implicit assumption that the extension
didn't have any additional dots in it. This allowed it to consolidate
"*.java" || "*.scala" into ExtensionFilter("java", "scala"). This
optimization does not really work for more complicated extensions so I
added a check that the extension doesn't have any dots in it.

### DIFF
--- a/io/src/main/scala/sbt/io/NameFilter.scala
+++ b/io/src/main/scala/sbt/io/NameFilter.scala
@@ -333,10 +333,11 @@ object GlobFilter {
     else {
       val parts = expression.split("\\*", -1)
       parts match {
-        case Array("", ext) if ext.startsWith(".") => new ExtensionFilter(ext.drop(1))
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/241/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #243, opened 2019-05-31
Fix directory listing for files with special characters

eatkins:
Listing directories that contained files whose names contained special
characters, like "^M", was broken on windows. This was because the
swoval library was using the utf-8 encoding instead of the native java
encoding which uses two byte words for characters.

The relevant swoval PR is https://github.com/swoval/swoval/pull/130.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -11,6 +11,6 @@ object Dependencies {
   val scalatest = "org.scalatest" %% "scalatest" % "3.0.8-RC2"
   val jna = "net.java.dev.jna" % "jna" % "4.5.0"
   val jnaPlatform = "net.java.dev.jna" % "jna-platform" % "4.5.0"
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/243/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I verified on my windows vm that running sbt with an io version with a dependency on swoval 2.1.2 would not crash when it encountered files with special characters in them. I also added a test in swoval for a file with '^M'.

Fixes https://github.com/sbt/sbt/issues/4690

eed3si9n (approved):
Thanks!

PR #252, opened 2019-07-30
Pathfinder ordering

eatkins:
Ref: https://github.com/sbt/sbt/issues/4909


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/252/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
nice. I wonder if this would resolve the last bits of mysterious behaviors.

eatkins:
I don't expect this to make a huge difference in general since I think most usages of PathFinder use extension filters which typically filter out directories anyway. It can't hurt though.

PR #253, opened 2019-08-01
Exclude hidden files by default in globs

eatkins:
This came up while I was writing documentation for Glob. The glob `baseDirectory.value / *` should  exclude hidden files by default. This is what normal shell globs do. To work around, if a user really wants hidden files to be returned, they can write `baseDirectory.value / "regex:\\..*`.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/Glob.scala
+++ b/io/src/main/scala/sbt/nio/file/Glob.scala
@@ -20,6 +20,7 @@ import sbt.nio.file.RelativeGlob.{ PathComponent, SingleComponentMatcher }
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.Properties
+import scala.util.matching.Regex
 
 sealed trait Glob {
 
@@ -37,12 +38,15 @@ object Glob {
     ordering.compare(left, right)
   def apply(file: File): Glob = if (file.isAbsolute) Root(file.toPath) else apply(file.toString)
   def apply(file: File, relative: String): Glob = Glob(file, RelativeGlob.parse(relative))
+  def apply(file: File, relative: Regex): Glob = Glob(file, s"regex:${relative.regex}")
   def apply(file: File, relative: RelativeGlob): Glob = Glob(file.toPath) / relative
   def apply(path: Path): Glob = if (path.isAbsolute) Root(path) else apply(path.toString)
   def apply(path: Path, relative: RelativeGlob): Glob = Glob(path) / relative
   def apply(path: Path, relative: String): Glob = Glob(path) / relative
+  def apply(path: Path, relative: Regex): Glob = Glob(path, s"regex:${relative.regex}")
   def apply(glob: String): Glob = {
-    val parts = splitter(glob)
+    val (regex, tail) = if (glob.startsWith("regex:")) (true, glob.drop(6)) else (false, glob)
+    val parts = splitter(tail)
     @tailrec def fullGlob(path: Path, rest: List[String]): Glob = {
       rest match {
         case component :: tail if !hasMeta(component) => fullGlob(path.resolve(component), tail)
@@ -50,6 +54,8 @@ object Glob {
         case _                                        => Pattern(path, RelativeGlob(rest: _*))
       }
     }
+    def addRegexPrefix(parts: List[String]): List[String] =
+      if (regex) parts.map(p => s"regex:$p") else parts
     parts match {
       case h :: rest if !hasMeta(h) =>
         val base = if (isWin) {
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/253/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
What does plain NIO do in this regard?

eatkins:
What do you mean? Does Files.walkFileTree return hidden files? If so, yes, it returns them. The java PathMatcher also includes them: https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-. However, I think in the build tool context, it does not make sense to include hidden files. The `excludeFilter` seems to exist almost exclusively to filter out hidden files. Shell globs like bash and zsh do not ~~exclude~~ (whoops, meant include) them. I am pretty sure that the only reason that PathMatcher doesn't exclude them is because `.` files aren't semantically hidden on windows.

The only use case I've seen for hidden files in sbt is for singe configuration files, e.g. `.sbtrc`. For those special files, I think it's fine to force the user to explicitly include them. Just like how I have to do `ls .foo` to see the hidden file `.foo`. It is excluded if I do `ls *`.

eatkins:
The more I think about this, I'm not 100% sure what to do. `Files.isHidden` is expensive on windows. It's similar to calling stat. I looked at Mill and they exclude all files starting with `.` (for scala and java sources at least). Gradle has a weird set of auto exclusions that they inherit from ant. It also seems difficult to change the rules: https://github.com/gradle/gradle/issues/2986
https://discuss.gradle.org/t/filetree-and-default-excludes/7406
http://ant.apache.org/manual/dirtasks.html#defaultexcludes

My thinking is that the Glob api should specify what it _includes_ and there should be little need for exclusions. The issue is that if we don't exclude hidden files by default, then either people will keep having to define `excludeFilter` in addition to `inputFiles` or they'll have to do something like:
```
fileInputs += baseDirectory.value.toGlob / "src" / "main" / ** / "!.*.scala.twirl"
```
That's not the end of the world, but it's somewhat noisy.

But windows is a problem because I don't really want to call Files.isHidden all over the place because that could cause real slowdowns. On the other hand, I think it's probably ok to say that sbt source files just shouldn't start with `.` as long as it's well-documented.

eed3si9n:
> What do you mean? Does Files.walkFileTree return hidden files? If so, yes, it returns them.

Yeah, that's what I meant. I just wanted to understand what they did.

This is kind of an interesting collision of cultural baggages in a way. We think of dotfiles as "hidden", but in a way, it's a naming convention adopted by UNIX/macOS people.

I'm guessing that NIO probably can't do it because as Java they have to be OS-agnostic, including Windows.

For sbt, it surely feels like most environment we cross across are macOS or Linux, but also in most situations sbt usually tries to keep Windows-parity. At the very least, I think there should be Scaladoc explaining the fact that `"."`-prefixed files are hidden, and how to recover them.

eatkins:
Yeah. I am writing website docs for glob right now and was thinking that there are a lot of missing scaladocs in Glob since I rushed that out back in May.

eatkins:
> This is kind of an interesting collision of cultural baggages in a way. We think of dotfiles as "hidden", but in a way, it's a naming convention adopted by UNIX/macOS people.

Yeah. The inverse way of looking at it though is that it's a good reason for excluding dotfiles even on Windows since they have different semantics on UNIX/macOS. In general, hidden files cause problems for portability since Files.isHidden could return something different depending on what platform is being used. Not that this is likely, but using `excludeFilter := HiddenFileFilter`, a windows user could add a dotfile and break the build for a linux user.

eatkins:
Hmm. I just realized that jar files sometimes contain hidden files. Maybe it is best to just include dot files alone and rely on exclude filter for sources. I guess it's well worth sleeping on. I still want to make the regex changes and need to add scaladocs anyway.

eed3si9n:
That reminds me that io 1.2.x's HiddenFilter says:

```scala
/** A [[FileFilter]] that selects files that are hidden according to `java.io.File.isHidden` or if they start with a dot (`.`). */
object HiddenFileFilter extends FileFilter {
  ...
}
```

sort of what you're saying, but except the Scaladoc is a lie and/or a case of a missing-comma because the implementation is actually:

```scala
/** A [[FileFilter]] that selects files that are hidden according to `java.io.File.isHidden` or if they start with a dot (`.`). */
object HiddenFileFilter extends FileFilter {
  def accept(file: File) = file.isHidden && file.getName != "."
}
```

It's just calling `File#isHidden`.

eatkins:
Yeah. I'm guessing that Mark Harrah wasn't doing a lot of windows testing in https://github.com/sbt/io/commit/32e87123b9d07e1add0ebb4f23977b95e21b4051.

eatkins:
That implementation is kind of weird. There is a certain logic to it because a single '.' file represents the current directory, which isn't usually a hidden file. But it is inconsistent with the scaladoc and, for the most part, sbt is good about normalizing paths so that that case shouldn't be relevant. So I can't even tell what the intent was.

eatkins:
Closing in favor of #257.

PR #259, opened 2019-08-07
Path filter

eatkins:
This adds the `PathFilter` trait, which is effectively `(Path, FileAttributes) => Boolean` and is used with the `Glob`/`fileTreeView` dsl similarly to the `sbt.io.FileFilter` for the `PathFinder` dsl.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/PathFilter.scala
+++ b/io/src/main/scala/sbt/nio/file/PathFilter.scala
@@ -32,10 +32,29 @@ private[sbt] trait LowPriorityPathFilter {
   /**
    * Converts a glob string to a [[sbt.nio.file.PathFilter]].
    * @param glob the glob string to convert to a filter, e.g. "**<code>/</code>*.scala"
-   * @return the parsed glob. May throw an exception if the glob string can not be parsed into a
+   * @return the [[PathFilter]] corresponding to the parsed [[Glob]]. May throw an exception if the
+   *         glob string can not be parsed into a [[Glob]].
+   */
+  implicit def stringToPathFilter(glob: String): PathFilter = new GlobPathFilter(Glob(glob))
+
+  /**
+   * Converts a [[Glob]] to a [[PathFilter]]. The [[PathFilter.accept]] method will ignore the
+   * [[FileAttributes]] parameter and return true if the `path` parameter is accepted by the
+   * input glob.
+   * @param glob the glob string to convert to a filter, e.g. "**<code>/</code>*.scala"
+   * @return the [[PathFilter]] corresponding to the parsed [[Glob]]. May throw an exception if the glob string can not be parsed into a
    *         [[Glob]].
    */
-  implicit def stringToPathFilter(glob: String): PathFilter = Glob(glob)
+  implicit def globToPathFilter(glob: Glob): PathFilter = new GlobPathFilter(glob)
+  private class GlobPathFilter(private val glob: Glob) extends PathFilter {
+    override def accept(path: Path, attributes: FileAttributes): Boolean = glob.matches(path)
+    override def toString: String = glob.toString
+    override def equals(obj: Any): Boolean = obj match {
+      case that: GlobPathFilter => this.glob == that.glob
+      case _                    => false
+    }
+    override def hashCode: Int = glob.##
+  }
 }
 object PathFilter extends LowPriorityPathFilter {
 
@@ -47,12 +66,7 @@ object PathFilter extends LowPriorityPathFilter {
    */
   def apply(globs: Glob*): PathFilter = if (globs.isEmpty) NoPass else new GlobPathFilter(globs)
 
-  /**
-   * Provides extension methods for combining or negating [[PathFilter]] instances or
-   * or other filter types that can be safely converted (see [[sbt.io.DirectoryFilter]]
-   * and [[sbt.io.HiddenFileFilter]]).
-   */
-  implicit class Ops(val pathFilter: PathFilter) extends AnyVal {
+  private[file] trait PathFilterExtensions extends Any {
 
     /**
      * Combines this filter with a [[sbt.nio.file.PathFilter]] to produce a combined filter
@@ -61,6 +75,31 @@ object PathFilter extends LowPriorityPathFilter {
      * @return the [[sbt.nio.file.PathFilter]] representing both filters combined by the `&&`
      *         operation
      */
+    def &&(other: PathFilter): PathFilter
+
+    /**
+     * Combines this filter with a [[sbt.nio.file.PathFilter]] to produce a combined filter
+     * that returns true either if this or the other [[sbt.nio.file.PathFilter]] accept the path.
+     * @param other the other [[sbt.nio.file.PathFilter]]
+     * @return the [[sbt.nio.file.PathFilter]] representing both filters combined by the `&&`
+     *         operation
+     */
+    def ||(other: PathFilter): PathFilter
+
+    /**
+     * Creates a new [[sbt.nio.file.PathFilter]] what accepts a `(Path, FileAttributes)` pair only
+     * if this filter does not accept it.
+     * @return the negated [[sbt.nio.file.PathFilter]] corresponding to this filter
+     */
+    def unary_! : PathFilter
+  }
+
+  /**
+   * Provides extension methods for combining or negating [[PathFilter]] instances or
+   * or other filter types that can be safely converted (see [[sbt.io.DirectoryFilter]]
+   * and [[sbt.io.HiddenFileFilter]]).
+   */
+  implicit class Ops(val pathFilter: PathFilter) extends AnyVal with PathFilterExtensions {
     def &&(other: PathFilter): PathFilter = pathFilter match {
       case NoPass  => NoPass
       case AllPass => other
@@ -71,14 +110,6 @@ object PathFilter extends LowPriorityPathFilter {
           case g       => new AndPathFilter(f, g)
         }
     }
-
-    /**
-     * Combines this filter with a [[sbt.nio.file.PathFilter]] to produce a combined filter
-     * that returns true either if this or the other [[sbt.nio.file.PathFilter]] accept the path.
-     * @param other the other [[sbt.nio.file.PathFilter]]
-     * @return the [[sbt.nio.file.PathFilter]] representing both filters combined by the `&&`
-     *         operation
-     */
     def ||(other: PathFilter): PathFilter = pathFilter match {
       case NoPass  => other
       case AllPass => AllPass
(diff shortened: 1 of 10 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/259/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/nio/file/PathFilter.scala
```diff
@@ -0,0 +1,259 @@
...
+  /**
+   * Converts an instance of [[sbt.io.FileFilter]] to an [[sbt.nio.file.PathFilter]]. It will
+   * de-structure the [[sbt.io.FileFilter]] if possible to convert it to an equivalent, and
+   * possibly more efficient, [[PathFilter]].
+   *
+   * @param fileFilter the filter to convert
+   * @return the converted.
+   */
+  implicit def fromFileFilter(fileFilter: FileFilter): PathFilter = fileFilter match {
+    case sbt.io.HiddenFileFilter    => HiddenFilter
```
eed3si9n: Would it make sense to keep the names same, since `DirectoryFilter` is `DirectoryFilter`. It gets confusing when are doing the same thing but named slightly differently like `HiddenFileFilter` vs `HiddenFilter`, `NothingFilter` vs `NoPass `

eed3si9n (changes requested):
overall LGTM, but I have a question about naming.

eatkins:
Yeah. I hear that. I was trying to avoid using the same names for things but I couldn't think of a different name for DirectoryFilter.  I specifically didn't like `HiddenFileFilter` because it can be read as Hidden**FileFilter** which makes it sound like its an instance of `sbt.io.FileFilter`.
 One alternative would be to do move things into the `PathFilters` companion object and then it could be something like, `PathFilters.IsDirectory`, `PathFilters.IsHidden`, etc.

Alternatively, I could make a new package, `sbt.nio.file.filters` and then similarly it would be `sbt.nio.file.filters.IsDirectory`, `sbt.nio.file.filters.IsHidden`, etc.

Or we could just reuse the names.

eed3si9n:
I prefer `HiddenFileFilter` since it's normally already in the scope, so I don't have to think about the package name, and the meaning is clear to me. Maybe it should move to `sbt.nio` instead of `sbt.nio.file`, so we can write `nio.HiddenFileFilter`.

eatkins:
The existing `DirectoryFilter` and `HiddenFileFilter` can be used in a build.sbt file without import:
```
scalaSources := fileTreeView.value.list(baseDirectory.value.toGlob / ** / "*.scala", -HiddenFileFilter)
```
because I added an implicit conversion from `FileFilter` to `PathFilter` (so in many cases, no import is necessary). Does that affect your thinking?

eed3si9n:
> because I added an implicit conversion from FileFilter to PathFilter (so in many cases, no import is necessary). Does that affect your thinking?

I guess that's a workaround for the package names, but my opinion hasn't changed about the filter names, esp if the aspiration is that this becomes the standard in the future.

eatkins:
Right, so to me it still comes down to what is the best name for the filters regardless of where they are defined. I chose names like, `AllPass` and `NoPass` because having `Filter` in the name read like Hungarian notation. I think part of the challenge is that because sbt implicitly imports `sbt.io._` into the default scope in build.sbt files, they needed to have Hungarian names since there was no context for the import.

In the context of sbt 1.x, I think it's best to avoid name conflicts because if the nio filters have the same names as the io filters, then wildcard imports can break things. Moreover, I have some quibbles with some of the existing filter names. `HiddenFileFilter` is ambiguous because it isn't clear if it means that it is `FileFilter` or that it only applies to hidden files, but not directories. I didn't like `NothingFilter` because `Nothing` is the bottom type in scala so it could actually be interpreted as a filter that accepts everything (yes, I'm being pedantic). 

The more I think about it, I am becoming partial to `IsX` names like: `IsDirectory`, `IsRegularFile`, `IsHidden` and `NotHidden` (this is so common that I think it deserves a special name, though I'd be fine with users having to write `!IsHidden`). I don't care that much about `AllPass` and `NoPass` because it's rare that the user would specify either of those but I still think it would be good to not alias the io filter names.

The counterargument to having different names for nio filters is that if we drop the `import sbt.io._` from build.sbt in sbt 2.x, then a number of sbt 1.x builds will break. This could be handled through deprecation though.

eed3si9n:
I have no doubt there could have been a better naming scheme if it were a green field project, but the cognitive load and migration cost is not worth it to rename them in my opinion.

eatkins:
This isn't a rename though. These are completely different apis. The main reason for the nio package is to distinguish the replacement apis from the old apis. If you look at `java.io._` vs. `java.nio.file._`, there are numerous apis that do effectively the same thing but have different names or take different forms, e.g. `Files.getLastModifiedTime(path)` vs. `file.lastModified`. I think having the filters use the same names in io and nio _increases_ cognitive load because I can't tell what kind of filter it is unless it is qualified with the package, e.g. `nio.DirectoryFilter` vs `io.DirectoryFilter`.

At some point users have to change their builds to use the new apis.  There are three scopes in sbt main in which `includeFilter` and `excludeFilter` are used: `unmanagedSources`, `unmanagedResources` and `unmanagedJars`. A while back, I migrated `unmanagedSources` and `unmanagedResources` to use the nio apis internally. The conversions from io -> nio exist solely so that builds that set the include or exclude filters in those scopes, e.g.
```
unmanagedResources / excludeFilter := NothingFilter
```
will continue to work as they currently do.

Otherwise there is no reason to mix and match io and nio apis. In fact, I now think that the conversions should be made `private[sbt]`. All of the new apis are opt-in and have documentation (though I have to add a section about filtering to both the glob and file tracking docs).  In sbt main, I'm introducing two new settings `inputPathFilter` and `outputPathFilter`, that complement `fileInputs` and `fileOutputs`. These are analogous to `includeFilter` and `excludeFilter`. The only reason an sbt 1.3.0 user would encounter the nio filters is if they are writing an incremental task or if they explicitly use the `FileTreeView.list` api.

If we decide to make `sbt.nio` default in 2.x and it turns out the migration friction is too high, we can always either alias or redefine the filters in the `sbt.nio` package (similar to how `WatchSource` is an alias for `sbt.internal.io.Source`, we could have `type DirectoryFilter = sbt.nio.file.IsDirectory` in the `sbt.nio` package object). This is part of why I didn't put the filters top level in `sbt.nio`: it gives us more flexibility moving forward.

eed3si9n:
My overall point is that if two things are designed to perform a same task, but in different contexts, they should be named the same. On the other hand, if two things are meant to do different things, they should NOT share the same name. I think users are good at remapping the context occasionally (e.g. `map` in `Option` vs `map` in `List`). 

>  If you look at java.io._ vs. java.nio.file._, there are numerous apis that do effectively the same thing but have different names or take different forms

I don't that's good either. Though in nio's case they may argue that they had to do that because of vararg options pattern.

> If we decide to make sbt.nio default in 2.x and it turns out the migration friction is too high, we can always either alias or redefine the filters in the sbt.nio package

I guess I am front loading the discussion because I don't trust myself to remember to think about this when it happens, and also we are introducing Glob now.

eatkins:
Given those concerns, I'd rather make the nio filters `private[sbt]` than mirror the io names. That gives us the ability to re-use the names down the line if we wish, but eliminates the possibility of forgetting about the discussion (we could also add a link to this PR discussion and/or scaladoc outlining the issues to PathFilter.scala to ensure that the discussion is not lost).

We could add an implicit conversion from `FileFilter => PathFilter` in the `sbt.nio.file.syntax` package object. Usage would then be:
```
inputPathFilter := (_, attributes) => !attributes.isDirectory
```
or
```
import sbt.nio.file.syntax._
inputPathFilter := -DirectoryFilter
```
This would force the issue if we wanted to bring the nio filters into the default sbt context. My biggest issue with sharing names is that it isn't possible to make the nio apis first class. Say we shared all the names and put them in the `sbt.nio` package, then in a build.sbt:

```
import sbt.nio._
// won't compile because there are two DirectoryFilters in scope
inputPathFilter := DirectoryFilter
```
so either I have to do

```
import sbt.nio.{ DirectoryFilter => DirFilter, _ }
inputPathFilter := DirFilter
```
or I have to name every import from sbt.nio that I want.

In fact, the reason I changed the name of `HiddenFileFilter` was because when I introduced it, one of the io tests stopped compiling because of the name conflict.

It also makes it impossible to support having both io filters and nio filters in the default sbt scope (by which I mean adding `import sbt.nio._` to the build.sbt auto-imports). If they have no conflicting names, then we could import both. That being said, I have mixed feelings about the auto-imports in build.sbt because I think it's a little weird that in project/*.scala files you end up having to manually import everything. While auto import makes the example build.sbt files nice and short, eliding the imports may ultimately add more confusion than it's worth.

eed3si9n:
> Given those concerns, I'd rather make the nio filters private[sbt] than mirror the io names.

Sounds good to me.

If 

```scala
import sbt.nio.file.syntax._
inputPathFilter := -DirectoryFilter
```

can punt the issue, then that sounds like a good middle ground?

eatkins:
It occurred to me that I could just make `HiddenFileFilter`, `DirectoryFilter` and (the newly added) `RegularFileFilter` inherit both FileFilter and PathFilter. I still don't love the mix and match aspect of it, but it seems pretty workable. (see https://github.com/sbt/io/pull/259/commits/9ab00c57bb54d76c3785bd8da5090a1e1b9be061)

The reality is that none of these filters are all that commonly used. I did a github search for includeFilter, excludeFilter, HiddenFileFilter and DirectoryFilter. I found that DirectoryFilter was almost never used (it showed up in 4 *.sbt files). HiddenFileFilter was used about 130 times. The includeFilter and excludeFilter settings were used about 1000 and 700 times respectively. The includeFilter almost always took the form:
```
includeFilter := "*.txt" || "*include*"
```
The excludeFilter was typically something like:
```
excludeFilter := HiddenFileFilter || "*_test.scala" || "*skip*"
```
Often the `HiddenFileFilter` was not included and they were just using path name based filtering.

Using the nio apis, the above could be written:
```
inputIncludePathFilter := ** / "{*.txt,*include*}"
```
or
```
inputIncludePathFilter := "**/{*.txt,*include*"}
```
and
```
inputExcludePathFilter := HiddenFileFilter || (** / "{*_test.scala,*skip*"})
```
or
```
inputExcludePathFilter := HiddenFileFilter || (** / "*_test.scala") || (** / "*skip*")
```
or
```
inputExcludePathFilter := HiddenFileFilter.toNio || "**/{*._test.scala,*skip*}"
```
(the `toNio` is necessary because `HiddenFileFilter || "**/{*.test.scala,*skip*}"` is interpreted as `(HiddenFileFilter: FileFilter).||("{*.test.scala,*skip*}": NameFilter)`)

eed3si9n (approved):
LGTM

PR #263, opened 2019-08-13
Don't require absolute paths in toGlob calls

eatkins:
While working on a plugin, I discovered that calling toGlob on a
relative path would cause an exception to be thrown. The workaround was
to call Glob(path.toString). There are valid uses cases for converting
paths to relative globs though, so I think it should be supported.

I added special handling for empty paths in the conversion because if
paths are programatically generated they may be empty. For example, in
`val glob = foo.resolve(bar).toGlob / **` , then the base path name
might be empty if `foo == bar`. This would cause the glob to pretty much
never work because paths rarely contain an empty path.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/Glob.scala
+++ b/io/src/main/scala/sbt/nio/file/Glob.scala
@@ -274,6 +274,10 @@ object Glob {
   private object Root {
     implicit val ordering: Ordering[Root] = Ordering.by(_.root)
   }
+  private[file] final case object Empty extends RelativeGlob {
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/263/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #264, opened 2019-08-15
Apply DeMorgan's rule to not filters

eatkins:
I noticed that while the unary `!` operator worked correctly in sbt, the
toString looked wrong because NotPathFilter toString was defined as:
`override def toString: String = !filter`
When the filter was an || filter, this turned into something like
!DirectoryFilter || HiddenFileFilter
which was confusing because it actually represented
!(DirectoryFilter || HiddenFileFilter)

When I thought about it, I realized it would be nicer to just apply
DeMorgan's rules in the NotFilter. The filter then becomes:
!DirectoryFilter && !HiddenFileFilter
which I think is easier to read anyway.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/PathFilter.scala
+++ b/io/src/main/scala/sbt/nio/file/PathFilter.scala
@@ -122,12 +122,18 @@ object PathFilter extends LowPriorityPathFilter {
           case g       => new OrPathFilter(f, g)
         }
     }
-    def unary_! : PathFilter = pathFilter match {
-      case AllPass     => NoPass
-      case NoPass      => AllPass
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/264/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #266, opened 2019-08-21
Fix stack overflow in Glob ordering

eatkins:
This issue was reported in https://github.com/sbt/sbt/issues/4970. The
isssu was that the glob ordering had an infinite loop in it when
a Root glob was compared to a FullFileGlob (which are created to convert
directory + fileFilter to Glob for legacy support). The reason this
hadn't been seen is because the play plugin is one of the few places
that creates sbt.internal.io.Source from single paths, which is from where
the Root globs (which are rarely a part of the watch globs) are coming.
This seems to be a side effect of play putting assets in the source
directory:

```scala
watchSources ++= {
    ((sourceDirectory in Compile).value ** "*" --- (sourceDirectory in Assets).value ** "*").get
},
```

is the setting that triggered the issue.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/Glob.scala
+++ b/io/src/main/scala/sbt/nio/file/Glob.scala
@@ -262,11 +262,13 @@ object Glob {
       case Root(leftRoot) =>
         right match {
           case Root(rightRoot) => leftRoot.compareTo(rightRoot)
+          case _: FullFileGlob => -1
           case _               => -compare(right, left)
         }
       case FullFileGlob(leftBase, _, _) =>
         right match {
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/266/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #267, opened 2019-08-30
Fix and comprehensively test glob ordering

eatkins:
The Ordering[Matcher] implementation violated the general ordering
contract. https://github.com/sbt/sbt/issues/4970 was still present
with sbt 1.3.0-RC5. This commit fixes the contract of Ordering[Matcher]
so that it should be sound for all Matcher instances.

Using jacoco, I was able to identify which branches of
the matchers were not being hit. After this change, all of the branches
were hit. While I'm not sure what was specifically causing the issue in
sbt/4970, I suspect that somewhere was an AndFilter, OrFilter or
NotFilter. The conversion in Globs from AndFilter, NotFilter and OrFilter to
Matcher created AndMatcher, NotMatcher and OrMatcher. The contract
between these matchers was not sound in the Ordering[Matcher]
implementation. Instead of trying to sort these, I just throw up my
hands and say that they are uncomparable.

After this change, I have 100% coverage of all of the branches in the
ordering implementations. Note that the ordering is unspecified between
AndMatcher, NotMatcher and OrMatcher. This means that, for example,
{ **/!foo, **/(foo && bar) } and { **/(foo && bar), **/!foo }
are both possible orderings depending on the initial ordering of these
two globs. However,
{ **/bar/!foo, **/foo/!bar }
is the only possible ordering for those two globs. In other words,
ordering is preserved up until two components differ only by an
(And|Not|Or)Matcher.

Here are screenshots showing that all of the branches are covered in jacoco:
Screen Shot 2019-08-30 at 11 05 26 AM
Screen Shot 2019-08-30 at 11 04 52 AM
By comparison, here was the previous jacoco coverage:
Screen Shot 2019-08-30 at 11 09 37 AM
(whoops)

### DIFF
--- a/io/src/main/scala/sbt/nio/file/Glob.scala
+++ b/io/src/main/scala/sbt/nio/file/Glob.scala
@@ -262,13 +262,16 @@ object Glob {
       case Root(leftRoot) =>
         right match {
           case Root(rightRoot) => leftRoot.compareTo(rightRoot)
-          case _: FullFileGlob => -1
-          case _               => -compare(right, left)
+          case FullFileGlob(leftBase, _, _) =>
+            leftRoot.compareTo(leftBase) match {
+              case 0 => -1
+              case i => i
+            }
+          case _ => -compare(right, left)
         }
       case FullFileGlob(leftBase, _, _) =>
         right match {
           case FullFileGlob(rightBase, _, _) => leftBase.compareTo(rightBase)
-          case _: Root                       => 1
           case _                             => -compare(right, left)
         }
     }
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/267/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #268, opened 2019-09-04
Close stream used in FileTreeView.nio

eatkins:
We created a stream without closing it. This came up because I was
reading the release notes for metals 0.7.2 and they mentioned fixing a
similar bug in the release notes.

In general, the nio FileTreeView should not really be in use in sbt
1.3.0. It would only be used on 32 bit platforms running sbt. Because
FileTreeView is a new api, and I rolled back the default path finder
implementations, none of the downstream consumers should be affected.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/FileTreeView.scala
+++ b/io/src/main/scala/sbt/nio/file/FileTreeView.scala
@@ -56,8 +56,9 @@ object FileTreeView {
   val nio: FileTreeView.Nio[FileAttributes] = (path: Path) =>
     Retry(
       {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/268/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #273, opened 2019-10-07
Fix IO.getModifiedTimeOrZero for unnormalized paths

eatkins:
The fix for long path names on windows,
099aa4f3ced649b48a9cf6595fbdb5524b69cfa5, had the unintended side effect
of causing IO.getModifiedTimeOrZero to throw uncaught IOExceptions when
the path name was not normalized. Specifically, it would throw if the
path was empty or if it had relative components. To fix this, we try to
get the modified time but if the file is not found, we normalize the
path and make it absolute and retry.

This change was primarily made for windows, but it turns out that
IO.getModifiedTimeOrZero(file("")) also didn't work correctly. Naively,
I'd expect the empty path to be expanded to the program working
directory but it would actually just return 0 because the posix native
api did not handle empty strings either. Relative paths did work on
posix before this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
      env:
          CMD="mimaReportBinaryIssues scalafmtCheckAll headerCheck Test/headerCheck test whitesourceCheckPolicies doc"
          TRAVIS_SCALA_VERSION=2.12.8
-         TRAVIS_JDK=adopt@1.8.192-12
+         TRAVIS_JDK=adopt@1.8.0-222
      os: linux
      dist: trusty
      install:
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/273/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #274, opened 2019-10-18
Prefer JavaMillis starting from Java 10

slandelle:
Motivation:

JavaMillis is currently only preferred for Java 11+.
But bug that chopped off millis from Files.getLastModifiedTime was fixed in Java 10, see https://bugs.openjdk.java.net/browse/JDK-8177809

Modification:

Have jdkTimestamps evaluate to true for Java 10.

Result:

Milli doesn't use JNA on Java 10.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Milli.scala
+++ b/io/src/main/scala/sbt/internal/io/Milli.scala
@@ -344,6 +344,7 @@ object Milli {
   //
   // If the property "sbt.io.jdktimestamps" is set to anything other than
   // "false", disable native millisecond-accurate modification timestamps.
+  // millis were missing until Java 10, see https://bugs.openjdk.java.net/browse/JDK-8177809
   //
   private val jdkTimestamps = {
     System.getProperty("sbt.io.jdktimestamps") match {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/274/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

slandelle:
Side question: why do you prefer `java.nio.Files#getLastModifiedTime` over `java.io.File#lastModified`?

eatkins:
> Side question: why do you prefer java.nio.Files#getLastModifiedTime over java.io.File#lastModified?

Long term we want to move all of the sbt io apis from io -> nio. Even if the public api exposes io, we have started moving the implementations to largely use nio under the hood.

eatkins (approved):
LGTM.

PR #283, opened 2020-01-19
[1.3.x] Fix IO.getModifiedTimeOrZero for unnormalized paths

eed3si9n:
This is a backport of sbt/io#273
Fixes https://github.com/sbt/sbt/issues/5224

## original description

The fix for long path names on windows,
099aa4f, had the unintended side effect
of causing IO.getModifiedTimeOrZero to throw uncaught IOExceptions when
the path name was not normalized. Specifically, it would throw if the
path was empty or if it had relative components. To fix this, we try to
get the modified time but if the file is not found, we normalize the
path and make it absolute and retry.

This change was primarily made for windows, but it turns out that
IO.getModifiedTimeOrZero(file("")) also didn't work correctly. Naively,
I'd expect the empty path to be expanded to the program working
directory but it would actually just return 0 because the posix native
api did not handle empty strings either. Relative paths did work on
posix before this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,15 +33,16 @@ before_install:
   # adding $HOME/.sdkman to cache would create an empty directory, which interferes with the initial installation
   - "[[ -d $HOME/.sdkman/bin/ ]] || rm -rf $HOME/.sdkman/"
   - curl -sL https://get.sdkman.io | bash
-  - echo sdkman_auto_answer=true > $HOME/.sdkman/etc/config
+  - echo sdkman_auto_answer=true >> $HOME/.sdkman/etc/config
+  - echo sdkman_auto_selfupdate=true >> $HOME/.sdkman/etc/config
   - source "$HOME/.sdkman/bin/sdkman-init.sh"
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.3.x&zinc=1.3.x&io=pull/283/head&librarymanagement=1.3.x&util=1.3.x&website=1.3.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
This failed on Travis CI macOS job.

```
[info] - should handle rapid creation of many subdirectories and files *** FAILED ***
[info]   result was false (SourceModificationWatchSpec.scala:451)
```

I'm going to restart and see if it passes.

eed3si9n:
@eatkins I'm going to merge this.

eatkins:
My bad. The review request notification slipped through the cracks.

PR #285, opened 2020-01-21
Correctly exclude base from Path.allSubpaths result

eatkins:
It was reported in https://github.com/scala/scala/pull/8525 that sbt
does not correctly exclude the base directory when calling
Path.allSubpaths. This ended up causing jar files to be created with an
empty entry. The problem was in an incorrect equality check because the
compiler did not flag for us that
`PathFinder(base).globRecursive(filter).get()` returned `Seq[File]` but
that we were filtering the results by comparing a `File` to a `Path`.

### DIFF
--- a/io/src/main/scala/sbt/io/PathMapper.scala
+++ b/io/src/main/scala/sbt/io/PathMapper.scala
@@ -112,12 +112,10 @@ abstract class Mapper {
    * Selects descendants of `base` directory matching `filter` and maps them to a path relative to `base`.
    * `base` itself is not included.
    */
-  def selectSubpaths(base: File, filter: FileFilter): Traversable[(File, String)] = {
-    val path = base.toPath
+  def selectSubpaths(base: File, filter: FileFilter): Traversable[(File, String)] =
     PathFinder(base).globRecursive(filter).get().collect {
-      case f if f != path => f -> path.relativize(f.toPath).toString
+      case f if f != base => f -> base.toPath.relativize(f.toPath).toString
     }
-  }
 
   /**
    * return a Seq of mappings which effect is to add a whole directory in the generated package
--- a/io/src/test/scala/sbt/io/PathMapperSpec.scala
+++ b/io/src/test/scala/sbt/io/PathMapperSpec.scala
@@ -128,6 +128,13 @@ class PathMapperSpec extends fixture.FlatSpec with Matchers {
     mappings should be(empty)
   }
 
+  it should "not include the base directory" in { tempDirectory =>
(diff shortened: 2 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/285/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/io/PathMapper.scala
```diff
@@ -112,12 +112,10 @@ abstract class Mapper {
    * Selects descendants of `base` directory matching `filter` and maps them to a path relative to `base`.
    * `base` itself is not included.
    */
-  def selectSubpaths(base: File, filter: FileFilter): Traversable[(File, String)] = {
-    val path = base.toPath
+  def selectSubpaths(base: File, filter: FileFilter): Traversable[(File, String)] =
     PathFinder(base).globRecursive(filter).get().collect {
-      case f if f != path => f -> path.relativize(f.toPath).toString
+      case f if f != base => f -> base.toPath.relativize(f.toPath).toString
```
eed3si9n, line 117: I wonder if there could have been some sort of linting that could've caught this.

eed3si9n (approved):

io/src/main/scala/sbt/io/PathMapper.scala
eatkins: Yeah. I'm sort of surprised that I didn't get an unrelated type warning. From what I can tell, this seems baked into scala2: https://github.com/lampepfl/dotty/issues/1247. I wonder if there is a compiler plugin though that can catch this kind of thing.

som-snytt: It does feel like help should be nearby.
```
val f = new java.io.File("f"); val p = f.toPath; f match { case `p` => }
```
By coincidence, I just tweaked a [similar expression](https://github.com/scala/scala/pull/8598/commits/4af22b3267a3646dfd182fd196626aedd20bc716#diff-5896f2db23c2c8679f9b2bb44c18adc1R175) albeit different file and path. I tend to stick to `nio.file.Path` now. There is a gotcha in that `p.endsWith("f")` is not `String::endsWith`.

dwijnand:
No backport candidate flag?

eed3si9n:
Created the label.

PR #287, opened 2020-01-27
[1.3.x] Correctly exclude base from Path.allSubpaths result

eed3si9n:
This is a backport of https://github.com/sbt/io/pull/285

It was reported in https://github.com/scala/scala/pull/8525 that sbt
does not correctly exclude the base directory when calling
Path.allSubpaths. This ended up causing jar files to be created with an
empty entry. The problem was in an incorrect equality check because the
compiler did not flag for us that
`PathFinder(base).globRecursive(filter).get()` returned `Seq[File]` but
that we were filtering the results by comparing a `File` to a `Path`.

### DIFF
--- a/io/src/main/scala/sbt/io/PathMapper.scala
+++ b/io/src/main/scala/sbt/io/PathMapper.scala
@@ -112,12 +112,10 @@ abstract class Mapper {
    * Selects descendants of `base` directory matching `filter` and maps them to a path relative to `base`.
    * `base` itself is not included.
    */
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.3.x&zinc=1.3.x&io=pull/287/head&librarymanagement=1.3.x&util=1.3.x&website=1.3.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #299, opened 2020-08-06
Relax anti-entropy quarantine constraint

eatkins:
In sbt scripted tests, it was possible for files to trigger a few times
with the same timestamp which would cause the anti entropy file event
monitor to drop the event even if the anti-entropy period was set to
zero. This could be worked around with a negative anti-entropy period,
but this seems like a better fix.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/299/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #310, opened 2020-11-25
Avoid ConcurrentModificationException

eatkins:
I saw a ConcurrentModificationException in an sbt session while
traversing the observers map in Observers.onNext. This exception can
generally be avoided by using a ConcurrentHashMap rather than a
LinkedHashMap. In general, it shouldn't matter what order the
callbacks are evaluated so it feels ok to switch to the
ConcurrentHashMap even though we lose the insertion ordering.

In addition to changing observers to a ConcurrentHashMap, we can drop
the conversion of the observables map from a java.util.WeakHashMap to a
scala.collection.mutable.Map. There wasn't really any benefit to this
conversion since java 8 maps have the forEach method.

### DIFF
--- a/io/src/main/scala/sbt/internal/nio/Observers.scala
+++ b/io/src/main/scala/sbt/internal/nio/Observers.scala
@@ -12,13 +12,12 @@ package sbt.internal.nio
 
 import java.io.IOException
 import java.nio.file.{ Path => NioPath }
-import java.util
+import java.util.WeakHashMap
+import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.atomic.AtomicInteger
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/310/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #311, opened 2020-11-25
Retry sleep duration

eatkins:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/311/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

io/src/main/scala/sbt/internal/io/Retry.scala
```diff
@@ -24,7 +24,13 @@ private[sbt] object Retry {
   private[sbt] def apply[@specialized T](
       f: => T,
       limit: Int,
-      excludedExceptions: Class[_ <: IOException]*
+      excludedExceptions: Class[_ <: IOException]*,
+  ): T = apply(f, limit, 100, excludedExceptions: _*)
+  private[sbt] def apply[@specialized T](
+      f: => T,
+      limit: Int,
+      sleepDuration: Long,
```
eed3si9n: I feel like this should use Duration class, and have a val for 100ms so we don't have to allocate it each time.

eatkins: That seems like overkill for a private api with defaults.

eed3si9n: or at least change the param name to `sleepInMillis`

eatkins: done

eed3si9n (approved):

PR #319, opened 2021-05-30
Fix sbt bug #6453

eatkins:
A user pointed out that if two globs were specificed that looked like
a/b.txt and a/b/**, then FileTreeView.list(globs) would not return the
children in the subdirectory a/b. There is an optimization that
FileTreeView.list makes when handling paths that may have multiple globs
matching the same prefix. At the fork point (which in the example above
is the directory a), we save all of the files that we find in there so
that as the recursive algorithm proceeds, it doesn't necessarily need to
make an additional syscall to determine whether the next file it wants
to look at is a path or a directory. The problem here though was that we
incorrectly passed the parent (fork) path into the listPath method
rather than the path itself.

### DIFF
--- a/io/src/main/scala/sbt/nio/file/FileTreeView.scala
+++ b/io/src/main/scala/sbt/nio/file/FileTreeView.scala
@@ -372,7 +372,7 @@ object FileTreeView {
                   case Some((`p`, m)) =>
                     m.get(path) match {
                       case null               =>
-                      case a if a.isDirectory => listPath(p)
+                      case a if a.isDirectory => listPath(path)
                       case a                  => maybeAdd(path -> a)
                     }
                   case _ => listPath(path)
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=pull/319/head&librarymanagement=develop&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
I tested in a local sbt build that https://github.com/sbt/sbt/issues/6453 was fixed by this change.

eed3si9n (approved):
Thanks @eatkins!

PR #326, opened 2021-09-19
Add hardLink to CopyOptions and related copyFile/copyDirectory helpers

er1c:
This is interesting to "copy" files from a cache location to another working directory, without making actual IO copies of the files.  The [java API Files.createLink](https://docs.oracle.com/javase/7/docs/api/java/nio/file/Files.html#createLink(java.nio.file.Path,%20java.nio.file.Path)) uses a hardlink (rather than a symbolic link) to do this.  It also seems to be supported on Windows, so that's the extra plus.  Creating links for a directory are discouraged, so this does it as the individual file level.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ ThisBuild / version := {
   (sys.env.get("BUILD_VERSION") orElse sys.props.get("sbt.build.version")) match {
     case Some(v) => v
     case _ =>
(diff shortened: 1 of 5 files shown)


eed3si9n:
If something needs a hard link, probably best to be implemented on its own via NIO or something rather calling it copy.
Closing this for now.

PR #333, opened 2021-11-14
Scala 3

xuwei-k:


io/src/main/scala/sbt/internal/nio/FileEventMonitor.scala
```diff
@@ -239,10 +239,14 @@ private[sbt] object FileEventMonitor {
...
     }
-    @tailrec
     override final def poll(
         duration: Duration,
         filter: FileEvent[T] => Boolean
+    ): Seq[FileEvent[T]] = pollImpl(duration, filter)
+    @tailrec
+    private[this] final def pollImpl(
+        duration: Duration,
+        filter: FileEvent[T] => Boolean
```
xuwei-k, line 249: - https://github.com/lampepfl/dotty/issues/9054

io/src/test/scala/sbt/io/CopyDirectorySpec.scala
```diff
@@ -11,11 +11,10 @@
 package sbt.io
 
 import java.nio.file._
-import org.scalatest.diagrams.Diagrams
 import org.scalatest.flatspec.AnyFlatSpec
 import sbt.io.syntax._
 
-class CopyDirectorySpec extends AnyFlatSpec with Diagrams {
+class CopyDirectorySpec extends AnyFlatSpec {
```
xuwei-k, line 17: diagram assert does not work with Scala 3 (compile error) 😢

SethTisue: Is it a known bug in ScalaTest?

eed3si9n (approved):
LGTM

PR #340, opened 2022-12-04
Add FileFilter.nothing and FileFilter.everything

mdedetrich:
The PR and documentation should be self explanatory.  Only possible addition would be to add tests however it can be argued that the code makes it completely clear whats going on.


eed3si9n (approved):
Thanks, @mdedetrich!

PR #343, opened 2023-04-26
Bump swoval version

eatkins:
A metals user identified what appeared to be a performance regression introduced in v2.1.8 (https://github.com/swoval/swoval/issues/161). The new version, 2.1.10, attempts to restore the previous performance.


eed3si9n (approved):
Thanks @eatkins!

PR #350, opened 2023-09-11
Add JDK 17 to CI

mdedetrich:
Self explanatory. I only added it to the Linux CI since it seems that we are only testing multiple JDK's on Linux (probably to reduce CI queues). Can only improve this in a later PR


eed3si9n (approved):
Thanks!

PR #353, opened 2023-09-21
Bump swoval to v2.1.12

eatkins:
For somewhat unknown reasons the native directory lister was very slow on windows for repos with deeply nested directory trees. This new version of swoval defaults to nio apis for Windows while retaining the native listers on posix platforms (where as far as I know there are no performance issues).

See https://github.com/swoval/swoval/issues/161.


eed3si9n (approved):
Thanks @eatkins!

PR #368, opened 2024-05-06
fix macos CI

xuwei-k:
- https://github.com/actions/setup-java/issues/627
- https://github.com/actions/setup-java/issues/625

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
             java: 21
             jobtype: 2
           - os: macos-latest
-            java: 8
+            java: 11
             jobtype: 2
           - os: windows-latest
             java: 8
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -42,6 +42,8 @@ jobs:
         distribution: temurin
         java-version: "${{ matrix.java }}"
         cache: sbt
+    - if: ${{ matrix.os == 'macos-latest' }}
+      run: brew install sbt
```
eed3si9n: I don't know what brew does so I'd rather do something like:

```suggestion
      run: |
        mkdir -p "$HOME/bin/"
        curl -sL https://raw.githubusercontent.com/sbt/sbt/v1.9.9/sbt > "$HOME/bin/sbt"
        export PATH="$PATH:$HOME/bin"
        chmod +x "$HOME/bin/sbt"
```

eed3si9n (approved):
Thanks!

PR #372, opened 2024-10-02
remove jna dependency

xuwei-k:


eed3si9n:
ok. So the context here is that JNA was originally added to IO via https://github.com/sbt/io/pull/92 to use native code for millisecond precision timestamp. The native code workaround was removed in https://github.com/sbt/io/pull/367 because JDK-8177809 fixed the millisecond bug, and the workaround itself started to cause issue on Alpine. So in theory, IO itself no longer needs to depend on JNA? LGTM.

eed3si9n (approved):

PR #396, opened 2024-12-18
encode the entries in zip by passing in UTF-8 ctor

### LINKED ISSUE #7400: ZipOutputStream should encode using UTF-8
## problem

https://github.com/sbt/io/blob/2656359e9ce42dc7e901277775bd66e3e9ccd4f3/io/src/main/scala/sbt/io/IO.scala#L762

```scala
case None => (new ZipOutputStream(fileOut), "zip")
```

## expectation

We should encode the entries in zip by passing in UTF-8 to the ctor.

nathanlao:
close https://github.com/sbt/sbt/issues/7400

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -767,7 +767,7 @@ object IO {
             os.closeEntry()
 
             (os, "jar")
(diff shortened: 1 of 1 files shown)


eed3si9n (approved):
Thanks!

PR #397, opened 2024-12-29
fix: Retry to retry non-IOException

eed3si9n:
Ref https://github.com/sbt/sbt/issues/7873#issuecomment-2556030861

## Problem
Retry(...) only handles IOExceptions.

## Solution
This fixes it to handle NonFatal.

### DIFF
--- a/io/src/main/scala/sbt/internal/io/Retry.scala
+++ b/io/src/main/scala/sbt/internal/io/Retry.scala
@@ -34,19 +34,19 @@ private[sbt] object Retry {
       excludedExceptions: Class[? <: IOException]*,
   ): T = {
     require(limit >= 1, "limit must be 1 or higher: was: " + limit)
-    def filter(e: Exception): Boolean = excludedExceptions match {
+    def filter(e: Throwable): Boolean = excludedExceptions match {
       case s if s.nonEmpty =>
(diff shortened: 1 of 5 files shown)


io/src/main/scala/sbt/internal/io/Retry.scala
```diff
@@ -34,19 +34,19 @@ private[sbt] object Retry {
       excludedExceptions: Class[? <: IOException]*,
   ): T = {
     require(limit >= 1, "limit must be 1 or higher: was: " + limit)
-    def filter(e: Exception): Boolean = excludedExceptions match {
+    def filter(e: Throwable): Boolean = excludedExceptions match {
```
adpi2, line 37: `excludedExceptions` is still defined as `Class[? <: IOException]`. It should probably be a `Class[? <: Throwable]`.

eed3si9n: Yea. I didn't want to take chance with bincompat but I guess I can try and see what mima says.

eed3si9n: Addressed in e4da4c5e86e799d6041ba80a464e4a536e368b52

PR #398, opened 2024-12-29
fix: Retry on move

eed3si9n:
## Problem
move sometimes fails on Windows?

## Solution
Retry on move.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -1116,17 +1116,17 @@ object IO {
   /**
    * Moves the contents of `a` to the location specified by `b`.
    * This method deletes any content already at `b` and creates any parent directories of `b` if they do not exist.
-   * It will first try `File.renameTo` and if that fails, resort to copying and then deleting the original file.
-   * In either case, the original File will not exist on successful completion of this method.
    */
-  def move(a: File, b: File): Unit = {
-    if (b.exists)
-      delete(b)
-    createDirectory(b.getParentFile)
-    if (!a.renameTo(b)) {
-      copyFile(a, b, true)
-      delete(a)
-    }
+  def move(a: File, b: File): Unit = move(a.toPath(), b.toPath())
+
+  /**
+   * Moves the contents of `a` to the location specified by `b`.
+   * This method deletes any content already at `b` and creates any parent directories of `b` if they do not exist.
+   */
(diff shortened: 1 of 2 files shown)


adpi2 (approved):

retronym:
I think I found a root cause for the intermittent failure in the rename or move operation

When staight-to-jar compilation is being used, SBT tries to influence Scalac's policy for caching of instances of ZipFile by setting a system property (`scala.classpath.closeZip=true`)

However, a `Global` classloader had previously initialized `scala.reflect.io.ZipArchive.closeZipArchive` and is later reused for another round of compilation, it would still use the pooling of `ZipFile` instances which seems to break a later attempt to call move the JAR file.

I can get the same underlying error on a `Files.move` operation if I leaks an unclosed `ZipFile` on Windows.

```scala
object Test {
  def main(args: Array[String]): Unit = {
    val fPathString = "c:/zaugg/f.jar"
    val f = Paths.get(fPathString)
    val g = Paths.get("c:/zaugg/g.jar")
    Files.deleteIfExists(f)
    Files.write(g, "g".getBytes(StandardCharsets.UTF_8))
    val jarfsProvider = FileSystemProvider.installedProviders().asScala.find(_.getScheme == "jar").get
    val fs = jarfsProvider.newFileSystem(f, java.util.Map.of("create", "true"))
    fs.close()

    val zf = new java.util.zip.ZipFile(f.toFile)
    zf.getComment()
    // don't close it

    Files.move(f, g, StandardCopyOption.REPLACE_EXISTING)
    // Exception in thread "main" java.nio.file.FileSystemException: c:/msde/zaugg/f.jar -> c:/msde/zaugg/g.jar: The process cannot access the file because it is being used by another process
    // windows error code 32 under the hoods, "sharing error" usually for another process holding the file open, but
    // seemingly this can also be native code from the current process?
  }
}
```

It seems that prior to this PR, the approach of falling back to the `copyFile`/`delete` worked around the problem. In the system I am working on, reverting to the prior version of `io` fixed the regression.

A workaround is for the user to set this system property explicitly on the JVM.

I'd like to revisit the code in scalac and see if I can get rid of the pooling altogether.

retronym:
Oh, it turns out there is a straight forward resource leak in Scalac classpath implementation under `-Dscala.classpath.closeZip=true`

https://github.com/scala/bug/issues/13094

So my proposed workaround is not helpful.

PR #414, opened 2025-03-20
Small documentation fix to `RegularFileFilter`

matomer:
Changed the documentation to say that the filter selects normal files instead of saying that the filter selects directories.


eed3si9n (approved):
Thanks!

matomer:
Signed the CLA.

Issue #434, opened 2025-08-19
PollingWatchService fails on ignored filter test on Windows


eed3si9n:
See for example https://github.com/sbt/io/actions/runs/17044606710/job/48360607495?pr=431

```
[info] - should watch a directory for file creation *** FAILED ***
[info]   EventMonitorSpec.this.watchTest(parentDir, sbt.internal.io.EventMonitorSpec.pathFilter(created.toPath()), sbt.internal.io.EventMonitorSpec.contains(created.toPath()))(sbt.io.IO.write(created, "foo", sbt.io.IO.write$default$3, sbt.io.IO.write$default$4)) was false (SourceModificationWatchSpec.scala:65)
```

eed3si9n:
Seems to have been fixed by https://github.com/sbt/io/pull/439

PR #436, opened 2025-08-19
Try normalize() instead

eed3si9n:
Ref https://github.com/sbt/io/issues/434


eed3si9n:
This failed on both macOS and Windows.

PR #442, opened 2025-08-30
update mima and CI settings

xuwei-k:


build.sbt
```diff
@@ -88,83 +88,11 @@ val io = (project in file("io"))
...
-      exclude[DirectMissingMethodProblem]("sbt.io.AllPassFilter.unary_-"),
-      exclude[IncompatibleSignatureProblem]("sbt.io.PollingWatchService.pollEvents"),
-      exclude[IncompatibleSignatureProblem]("sbt.io.WatchService#WatchServiceAdapter.pollEvents"),
-      exclude[IncompatibleSignatureProblem]("sbt.io.WatchService.pollEvents"),
-    ),
+    mimaPreviousArtifacts := {
+      Set(
+        "1.10.5",
+      ).map(version => organization.value %% moduleName.value % version)
+    },
```
eed3si9n, line 95: I wonder if we should release IO 2.x so we can reset on mim like this.

PR #483, opened 2026-02-25
Use Iterable instead of Traversable

xuwei-k:
https://github.com/scala/scala3/blob/341bda0cbaa3ffd61ad77f4d7b4f777adb416df8/library/src/scala/package.scala#L51-L54


eed3si9n (commented):
I guess it's ok from bytecode perspective to expand type aliases. Do we need to check if this is a TASTy-compatible change?

eed3si9n (changes requested):
On hold for TASTy compat

PR #484, opened 2026-02-25
Update scalafmt version and config

xuwei-k:


xuwei-k:
- https://github.com/scalameta/scalameta/issues/4515

https://github.com/sbt/io/blob/57d33dd7fea4b50bf7b90b28f1aee11cbe5d78fa/io/src/main/scala/sbt/io/Path.scala#L549-L552


```
[error] (io / Compile / scalafmt) org.scalafmt.sbt.ScalafmtSbtReporter$ScalafmtSbtError: scalafmt: [3.10.7] io/src/main/scala/sbt/io/Path.scala:550: error: [dialect scala3 [with overrides]] `}` expected but `=>` found
[error]     val apply = if (errorIfNone) (a: File) => mapper(a) orElse fail(a) else mapper
```

eed3si9n:
Since we're sticking with 1.10.x series for sbt 2.x, `develop` is effectively discontinued as a family line?

PR #495, opened 2026-03-01
Use filterInPlace instead of retain

xuwei-k:
```
[warn] -- Deprecation Warning: /home/runner/work/io/io/io/src/main/scala/sbt/internal/nio/FileEventMonitor.scala:307:27 
[warn] 307 |      antiEntropyDeadlines.retain((_, deadline) => Deadline.now < deadline + retentionPeriod)
[warn]     |      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[warn]     |method retain in trait MapOps is deprecated since 2.13.0: Use filterInPlace instead
```

https://github.com/scala/scala3/blob/b5d15acdc4c6749a86eb32eb211e0ae08e7c20ce/library/src/scala/collection/mutable/Map.scala#L169-L170


eed3si9n (approved):
Thanks!

PR #496, opened 2026-03-01
Fix warning. use compiletime.uninitialized

xuwei-k:
```
[warn] -- Warning: /home/runner/work/io/io/io/src/main/scala/sbt/internal/io/DeferredWriter.scala:19:34 
[warn] 19 |  private var delegate0: Writer = _
[warn]    |                                  ^
[warn]    |`= _` has been deprecated; use `= uninitialized` instead.
[warn]    |`uninitialized` can be imported with `scala.compiletime.uninitialized`.
[warn]    |This construct can be rewritten automatically under -rewrite -source 3.4-migration.
```


eed3si9n (approved):
Thanks!

PR #497, opened 2026-03-02
Enable -Werror option

xuwei-k:


project/HouseRulesPluglin.scala
```diff
@@ -13,21 +13,10 @@ object HouseRulesPlugin extends AutoPlugin {
...
-          case _       => v == 12
-        }
-      })
-      .value
-      .toList,
+    scalacOptions ++= Seq(
+      "-Wconf:msg=Compiler synthesis of Manifest and OptManifest is deprecated:silent",
+      "-Wconf:msg=type Traversable in package scala:silent",
+    ),
+    scalacOptions += "-Werror",
```
eed3si9n: I'm +1 on enabling this, but also retain the ability to opt out using `-Dsbt.build.fatal=false` because of sbt-unidoc etc purpose to mash sbt/io with sbt/zinc and sbt/sbt.

eed3si9n (approved):
Thanks!

PR #540, opened 2026-08-05
perf: Deflate zip entries in parallel

hoangmaihuy:
**Disclaimer**: Developed with Claude Code and human review in the loop

Optimize performance found in https://github.com/sbt/sbt/issues/9554

`IO.zip` and `IO.jar` deflate every entry one after another on the calling thread, which leaves most of a multiprocessor idle during what is usually the dominant cost of building an archive.

This adds `IO.zipParallel` and `IO.jarParallel`: the same archives, byte for byte, with the deflating shared out over an `ExecutionContext` the caller names. `IO.zip` and `IO.jar` are untouched, so nothing changes for a caller that does not ask for it — `packageBin`, `packageSrc`, `packageDoc` and sbt 2.x's `Def.declareOutputDirectory` can all opt in.

## Benchmarks

M2 Max (12 cores, macOS arm64), OpenJDK 25.0.4, `ioBenchmarks/Jmh/run -wi 3 -i 5 -r 5 -f 1`. Every arm goes through `IO`, and each pair was checked to produce identical archives.

| corpus | sequential | parallel | |
|---|---|---|---|
| 4000 entries, 79 MB — `IO.zip` | 2412 ± 60 ms | 388 ± 16 ms | **6.2×** |
| 4000 entries, 79 MB — `IO.jar` | 2414 ± 55 ms | 404 ± 154 ms | **6.0×** |
| the same plus three 4 MB entries, 91 MB | 3070 ± 442 ms | 585 ± 48 ms | **5.2×** |
| one 72 MB entry | 3405 ± 122 ms | 3371 ± 81 ms | 1.0× |

One large artifact has nothing to overlap with, so it neither gains nor loses time. Nor does any of this depend on having cores to spare: with the deflating forced onto the writing thread the parallel writer comes in at 2316 ± 80 ms against 2412 ± 60 for zip, 2295 ± 56 against 2414 ± 55 for jar, and 3365 ± 144 against 3405 ± 122 for the large entry. The sequential mixed arm and the parallel jar arm are the noisy measurements, at ±442 ms and ±154 ms over five iterations.

Allocation per archive through `IO`, which the benchmark measures itself and fails the run over:

| | sequential | parallel |
|---|---|---|
| 4000 entries | 44.3–44.6 MB | 53.7–53.9 MB |
| one 72 MB entry | under 0.1 MB | 16.3 MB |

Holding an entry so another thread can deflate it is the one thing the sequential writer never does, which is the second row. Both bounds on it — how much may be in flight, and how large an entry may be before it is streamed instead — are fixed at 16 MB rather than scaled to the heap. On incompressible input under `-Xmx192m`, 300 × 1 MB peaks at 32 MB and one 300 MB at 38 MB, where disabling the streaming guard turns the second into an `OutOfMemoryError`; the same corpus completes at `-Xmx64m`. Nothing is retained across archives: 300 archives × 400 entries on JDK 8 holds flat at 2 MB heap and 233 MB rss, against a deliberately unclosed control that climbs to 363 MB.

## Using it

```scala
import sbt.io.IO.Implicits.zipContext  // commonPool
IO.zipParallel(sources, out, time)
```

Neither method defaults to a context: a default value on an implicit parameter is silently replaced by any `ExecutionContext` in scope, and where an archive deflates is worth one line at the call site. It sits in an object of its own because `IO` is wildcard-imported. Any context will do: the writing thread runs a deflation the context has not started rather than waiting on it, so one that is busy, single-threaded, or drops what it is given outright still completes the archive.

## Identical output

Callers depend on the exact bytes — artifact checksums, reproducible-build comparisons — so the output is what `ZipOutputStream` and `JarOutputStream` produced, down to which archives they refuse and in what words. Everything the JDKs disagree about is asked of a reference `ZipOutputStream` at run time rather than restated here, so a new JDK needs no change: an entry's MS-DOS timestamp cannot be recovered from `getTime()`, whether it gets an info-zip extra field is not observable through the public API, and reflection into `ZipEntry.xdostime` has been closed off since JDK 16. One divergence is deliberate — an entry opened after `finish` is refused, where the reference appends it past its own end record.

## Testing

397 tests, 112 of them zip. `ZipSpec` pins `IO.zip` and `IO.jar` against references driving the JDK writers longhand and writes every fixture a third time through the parallel pair; eight suites in `sbt.io.parallel` drive the writers directly, one per aspect. Byte-identity is self-obscuring, so the references run in-process on the running JDK, the held and streamed paths are asserted to give the same bytes, refusals are compared call for call, and the handovers a caller's context received are counted — an archive deflated somewhere else has the same bytes and would otherwise pass. A fuzzer over random entry counts, sizes, write chunking, level changes and hold thresholds compared 22,400 archives clean across JDK 8, 11, 21 and 25, four context shapes and two parallelism settings; it found the two divergences fixed here.

## Known limits

- **Peak heap is higher, and running out of it is an `OutOfMemoryError` rather than a `ZipException`** — 32–38 MB, fixed rather than scaled, so a heap below it fails where the reference would have completed.
- **Native memory is outside both bounds**: each entry in flight holds a `Deflater`, around 256 KB of zlib state.
- **What a context does with the work is the caller's problem.** A `RejectedExecutionException` reaches the caller and the archive can still be finished without that entry — `IO` cannot, since it leaves `writeFileAtomically` nothing to move into place. A context that drops work silently is deflated past on the writing thread rather than waited on. Where the writer does wait, on a deflation the context has already started, an interrupt ends that wait as an `InterruptedIOException` with the flag put back.
- Only those three are reachable through `IO`. The rest need a caller building its own entries: a stored entry whose compressed size this writer filled in is written differently the second time on 21 and later, one declaring a negative compressed size is refused with a different number in the message, and an entry opened after `finish` is refused.
- Not thread safe: one instance writes one archive from one thread.

## Compatibility

Green on Scala 2.12.20, 2.13.18 and 3.3.8, and on JDK 8, 11, 21 and 25: 396 pass on each, with one cancelled everywhere, where `setExtra` caps the extra field below what that case needs. MiMa is clean against 1.12.0 — `zipParallel`, `jarParallel` and `IO.Implicits` are new, `zip` and `jar` untouched, and the new classes are `private[sbt]`. The benchmark is a new `ioBenchmarks` module on sbt-jmh, aggregated so it keeps compiling but is never run by `test`.


eed3si9n:
Thanks for the contribution!

There was a similar contribution made in Zinc a while back that we might want to look into - https://github.com/sbt/zinc/pull/1326, which added `ParallelGzipOutputStream`.

If we can keep the `ZipOutputStream`-like interface, that might make the maintenance easier in the future. Also if we can parameterize `ParallelZipOutputStream` with `scala.concurrent.ExecutionContext` and parallelism, the caller can tweak the concurrency if we have to down the road.

hoangmaihuy:
Hi @eed3si9n, I've refactored the implementation to `ParallelZipOutputStream` and introduce its usage in new method `IO.zipParallel` so that sbt can safely use and test it in next milestone version.

`ParallelZipOutputStream` is only about 1k loc including comments, the rest are supportive files and unit tests. 
I've tried to reproduce identical byte output and behavior to `ZipOutputStream`, there is no good way to ensure that between different OS-es and JDK versions, so I use a probe `ZipOutputStream` for references, but the main idea is still parallel deflating entries using allocated buffers in given thread pool.

Hope this can help improving action cache performance, I'm glad to address any issues and comments 😄

io/src/main/scala/sbt/io/parallel/ParallelJarOutputStream.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+private[sbt] object ParallelJarOutputStream {
+
+  /** Whether an extra field already carries the jar magic id, walked as `JarOutputStream.hasMagic` walks. */
+  private[io] def hasMagic(extra: Array[Byte]): Boolean = {
+    // id before size, as the JDK does: a truncated trailing fragment of just the id counts as stamped
+    var i = 0
+    while (i + IdBytes <= extra.length) {
+      if (u16(extra, i) == JarMagicId) return true
```
eed3si9n: It's unusual for our codebase to have `var`, `while`, and `return` statements, one after the other - https://github.com/sbt/sbt/blob/develop/contributing-docs/03_coding_style.md

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -672,11 +674,49 @@ object IO {
...
+   * which is what `CompletableFuture` would have chosen anyway. Neither method defaults to it — a
+   * default on an implicit parameter is silently replaced by any context in scope rather than being
+   * the choice it looks like.
+   *
+   * Behind an object of its own, as `ExecutionContext.Implicits.global` is: `IO` is
+   * wildcard-imported, and an implicit `ExecutionContext` there is one every `Future` in the
+   * importing file would take silently or refuse as ambiguous against its own.
+   */
+  object Implicits {
+    implicit val deflateContext: ExecutionContext = ParallelZipOutputStream.commonPoolContext
```
eed3si9n: The term [deflate](https://en.wikipedia.org/wiki/Deflate) I think is a name for specific algorithm used by zip, so I'd rather not let it leak to our API. I suggest something like `zipContext`.

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -687,13 +727,36 @@ object IO {
...
+   *                    them deflate at a time is the context's business. One per processor by default
+   * @param ec where the deflating happens; [[Implicits.deflateContext]] is the one to import for
+   *           `commonPool`. It has to be a context this thread is not itself the whole of: the
+   *           writing thread waits on the deflating once enough entries are in flight
+   */
+  def zipParallel(
+      sources: Traversable[(File, String)],
+      outputZip: File,
+      time: Option[Long],
+      parallelism: Int = ParallelZipOutputStream.DefaultParallelism
```
eed3si9n: Please avoid default parameters, and define two overloads instead.

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
```diff
@@ -0,0 +1,917 @@
...
+  private var level = Deflater.DEFAULT_COMPRESSION
+  private var method = ZipEntry.DEFLATED
+  private var comment: Array[Byte] = Array.emptyByteArray
+
+  // held as blocks rather than one array grown into: doubling costs 2N allocated and N copied
+  private var blocks = new Array[Array[Byte]](InitialBlocks)
+  private var blockUsed = new Array[Int](InitialBlocks)
+  private var blockCount = 0
+  private var buffered = 0
+  private var holding = false
```
eed3si9n: Number of `var` here freaks me out a bit. `level` etc might be relatively safe since outside code would call it, but `blocks` feels like a source of a race condition if a bunch of methods are reading and writing into it without synchronization.

io/src/main/scala/sbt/io/parallel/ParallelJarOutputStream.scala
hoangmaihuy: I've refactored the code to use Scala syntax: tailrec, Option....

io/src/main/scala/sbt/io/IO.scala
hoangmaihuy: Renamed to `zipContext`

io/src/main/scala/sbt/io/IO.scala
hoangmaihuy: Created two overloads, one having `parallelism` params, one not having `parallelism` will use `IO.defaultParallelism`

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
hoangmaihuy: I've moved them inside a `HeldEntry` and double checked to make sure there is no race condition

eed3si9n:
Thanks for addressing the reviews!

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
```diff
@@ -0,0 +1,886 @@
...
+    def deflate(): Deflated =
+      deflateInto(entryDeflater, taken.blocks, taken.used, taken.bytes, into, window)
+    val task = new FutureTask[Deflated](() => deflate())
+    // a refused submission never reaches the queue, so its deflater is ended here rather than swept
+    try ec.execute(task)
+    catch {
+      case failed: Throwable =>
+        entryDeflater.end()
+        throw failed
+    }
```
eed3si9n, line 357: nit: Claude identified that this would leak buffer. It wants you to give the buffer back in `case`.

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
```diff
@@ -0,0 +1,886 @@
...
+
+    /** The deflation, run here if the context has not started it — `run` is a no-op once it has. */
+    def claimed(): FutureTask[Deflated] = {
+      task.run()
+      task
+    }
+
+    /** `cancel(false)` wins only on a task no thread started, so a running deflation keeps its deflater. */
+    def endDeflater(): Unit =
+      if (task.cancel(false) || task.isDone) deflater.end()
```
eed3si9n: Also per Claude

> `Deflater` isn't safe to `end()` while another thread is inside `deflate()` on it

```suggestion
    def endDeflater(): Unit = {
      task.cancel(false)
      try { task.run(); task.get() }
      catch {
        case interrupted: InterruptedException => Thread.currentThread().interrupt()
        case _: Throwable                      => ()
      } finally deflater.end()
    }
``

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
hoangmaihuy: Fixed, the refused entry now hands back both the blocks it was written into and the buffer it would have deflated to before rethrowing. I also added a test to verify this case.

io/src/main/scala/sbt/io/parallel/ParallelZipOutputStream.scala
hoangmaihuy: Reproduced in test, the sweep now runs the deflation where nothing else has and waits it out either way — the same thing claimed already does on a drain

eed3si9n (approved):
Thanks!

PR #552, opened 2026-08-30
 Add createFileAtomically and owner-only staging

kitbellew:
Two additional use cases for atomic file writing:
- create the file only if missing (such as portfile for sbt server)
- leave it readable only by the owner (token file for sbtn to auth to sbt server).

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -490,21 +490,20 @@ object IO {
     val staging = stagingFile.toPath
     touch(stagingFile)
 
+    def retry(func: => NioPath): NioPath = Retry(
+      func,
+      classOf[FileAlreadyExistsException],
+      classOf[AtomicMoveNotSupportedException],
+      classOf[UnsupportedOperationException]
+    )
+    def move(options: CopyOption*): NioPath = retry(Files.move(staging, toPath, options*))
+    def replaceFile(): NioPath =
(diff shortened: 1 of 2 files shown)


io/src/main/scala/sbt/io/IO.scala
```diff
@@ -475,36 +481,103 @@ object IO {
...
+  /**
+   * Like `writeFileAtomically`, except that it refuses a `to` that already exists and
+   * throws `FileAlreadyExistsException`.
+   *
+   * Where the filesystem creates hard links, one writer of many creates the file and the
+   * rest are refused, because a link fails with EEXIST. Where it does not, the refusal is
+   * the one `Files.move` gives: it reads the attributes of `to` and then renames, so two
+   * writers racing on a fresh `to` can both believe they created it.
+   */
+  def createFileAtomically[T](to: File, ownerOnly: Boolean = false)(write: File => T): T =
```
eed3si9n: Default parameter is difficult to keep future binary compatibility, so could you create a overload plz?

kitbellew:
@eed3si9n fyi helps with sbt/sbt#9689.

io/src/main/scala/sbt/io/IO.scala
kitbellew: done. didn't see the review before i made my other comment.

eed3si9n (approved):
Thanks!

eed3si9n:
https://github.com/sbt/io/releases/tag/v1.13.1 is out.

PR #553, opened 2026-08-30
Handling AccessDeniedException on project jar

ColOfAbRiX:
Part of the fix for [9651](https://github.com/sbt/sbt/issues/9651)

On Windows, Files.move with ATOMIC_MOVE can fail with AccessDeniedException when the target file is locked by another process (e.g., an sbt server or compiler holding a file handle). This causes intermittent build failures when writing jar files.

This change adds a catch clause for AccessDeniedException in [IO.writeFileAtomically()](vscode-webview://16530kufkd397s742d07t2qo7kbqfag08k8gt3r7cvrsbkdc0vrh/io/io/src/main/scala/sbt/io/IO.scala:478) that falls back to a stream copy via transferAndClose, followed by cleanup of the staging .tmp file.

Before: AccessDeniedException is retried 10 times by Retry (1 second of wasted retries), then propagates as a fatal error.

After: AccessDeniedException is caught after the Retry attempts exhaust, and the file is written via stream copy as a last resort.

### DIFF
--- a/io/src/main/scala/sbt/io/IO.scala
+++ b/io/src/main/scala/sbt/io/IO.scala
@@ -539,6 +539,10 @@ object IO {
       // Files.move with ATOMIC_MOVE possibly replaces and without has a race condition
       try retry(Files.createLink(toPath, staging)) // try this first
       catch {
+        case _: AccessDeniedException =>
+          transferAndClose(new FileInputStream(staging.toFile()), new FileOutputStream(to, false))
+          Files.deleteIfExists(staging)
+          ()
         case e @ (_: UnsupportedOperationException | _: IOException)
             if !e.isInstanceOf[FileAlreadyExistsException] =>
           move()


eed3si9n:
Hi, thanks for the contribution. It got conflicted with a PR that I merged earlier. Could you resolve it please?

io/src/main/scala/sbt/io/IO.scala
```diff
@@ -497,6 +497,10 @@ object IO {
       catch {
         case _: AtomicMoveNotSupportedException =>
           Retry(Files.move(staging, to.toPath(), StandardCopyOption.REPLACE_EXISTING))
+        case _: AccessDeniedException =>
+          transferAndClose(new FileInputStream(staging.toFile()), new FileOutputStream(to, false))
+          Files.deleteIfExists(staging)
+          ()
```
eed3si9n, line 545: I'm also not sure if it's a good idea to _not_ write files atomically when the function name specifically says it's writing file atomically. It might be a bit cumbersome, but it might be a good idea to implement another function like `writeFileAtomicallyMaybe` that does the fallback, and replace all call sites of `writeFileAtomically` to use that instead. cc @anatoliykmetyuk, @kitbellew

eed3si9n (changes requested):

io/src/main/scala/sbt/io/IO.scala
kitbellew: if the file is locked, can it be written to?

can it be deleted first? is a missing file better or worse than a half-written one?

kitbellew: i think the solution is to fix sbt/sbt#9651 because otherwise this looks like a way to undo a conscious decision made there: https://github.com/sbt/sbt/blob/952b82e4610841aec6be1241c81f9699722f850a/main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala#L169-L172

kitbellew: i think the solution is to fix sbt/sbt#9651 because otherwise this looks like a way to undo a conscious decision made there: https://github.com/sbt/sbt/blob/952b82e4610841aec6be1241c81f9699722f850a/main-command/src/main/scala/sbt/internal/classpath/ClassLoaderCache.scala#L169-L172

eed3si9n: As written in this PR's description, we did start seeing more AccessDeniedException on Windows when we implemented move-based atomic writing, so if we revert to file copying that's not atomic like in sbt 1.x, I'm guessing that we would see less of AccessDeniedException. Also due to sbtn and/or caching, we are also seeing the ill effect of half-written files too, so I wonder if we need to reconsider when to write files atomically, and if we don't write atomically use some second mechanism to communicate that the file(s) are ready for consumption?

kitbellew: i can see several options. client can retry. written out files can be versioned, with the client reading the highest version. could probably ask an llm for a suggestion, given some constraints.

ColOfAbRiX: To add a data point to this discussion, I've been using my patch locally for a week now and the AccessDeniedException completely disappeared. A log message shows that the issue is handled regularly and multiple times:
```
sbt:beesight> compile
[info] compiling 2 Scala sources to ...
[info] compiling 1 Scala source to ...
 *** Debugging the fix ***
[info] compiling 2 Scala sources to ...
[info] compiling 1 Scala source to ...
[success] elapsed time: 4 s, cache 88%, 66 disk cache hits, 9 onsite tasks
sbt:beesight> test
[info] compiling 5 Scala sources to ...
[info] Passed: Total 0, Failed 0, Errors 0, Passed 0
[info] No tests to run for beesightBenchmarks / Test / testQuick
 *** Debugging the fix ***
 *** Debugging the fix ***
[info] compiling 2 Scala sources to ...
```

I agree, in principle, that we should find a better solution than removing the atomic move. I'll work a little more and see if I can come up with anything better following your advice. Meanwhile, feel free to give me some more guidance

ColOfAbRiX:
I explored @kitbellew suggestions and they seem to involve many changes that I'm not comfortable doing myself. I could easily apply @eed3si9n suggestion to create a `writeFileAtomicallyMaybe` function, though.

Other suggestions I got from AI are:

**Files.copy as NIO fallback**

On Windows, URLClassLoader opens JARs with FILE_SHARE_READ | FILE_SHARE_WRITE but not FILE_SHARE_DELETE. This means Files.move fails (needs delete) but Files.copy(staging, toPath, REPLACE_EXISTING) succeeds — CopyFileEx overwrites content without deleting the file entry. If we add a fallback, using Files.copy is cleaner than manual FileInputStream/FileOutputStream stream copy: it's NIO, a single API call, and the idiomatic Java approach. The trade-off is that during the copy window, a concurrent reader could see a partially-written file — but in practice the copy is fast and ClassLoaders read lazily.

**Content-addressed JAR paths (never overwrite)**

sbt 2.x already uses HashedVirtualFileRef and FileConverter to map virtual file references to physical paths. If JARs were stored at content-addressed paths (e.g. target/out/store/<hash>.jar), they would never be overwritten — a new compile produces a new hash, hence a new path. The classpath reference is updated to point to the new hash. Old JARs are garbage-collected when no ClassLoader references them. This eliminates the AccessDeniedException entirely (no file is ever replaced while locked) and is principled (no silent degradation to non-atomic writes). The downside is scope: it requires changes to FileConverter, classpath resolution, and garbage collection.

anatoliykmetyuk:
Thanks for investigating this one! Closing as superseded by #558, released in 1.13.2.

I also verified on Windows that overwriting a JAR held open by a classloader fails with IO 1.13.1 and succeeds with 1.13.2. The fix is included in sbt 2.1.0-M1. Please let us know if you still encounter the problem with that version.

